PR #4, opened 2018-10-09
MODSOURCE-2 and MODSOURCE-4

Igor-Gorchakov:
- Create REST endpoint to return a list of "Source Change Jobs"
- Create REST endpoint to return a list of logs for "Source Change Jobs"

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,2 +1,15 @@
+## 2018-10-09 v1.0.0
+ * Add endpoints to retrieve source change jobs and logs for that jobs
+ * Implementation is currently stubbed
+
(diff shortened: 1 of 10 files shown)


dcrossleyau:
Those raml-cop complaints are because the "$ref" in each "*Collection.json" need the filename extension.

OleksiiKuzminov:
@Igor-Gorchakov plz change endpoints according to schema design document https://docs.google.com/document/d/1pDR6_jRw5jLKBn0rKWe-3zIak4eCFHbUs62X8jZdWck/edit

PR #9, opened 2018-11-13
MODSOURCE-8 Added GET endpoint for Result entity

KaterynaSenchenko:
* Added schema for Result entity
 * Added results_view to the db
 * Defined endpoint for GETting the results
 * Implemented the endpoint
 * Created API tests for the added endpoint

### DIFF
--- a/ramls/dto/result.json
+++ b/ramls/dto/result.json
@@ -1,0 +1,37 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Result DTO Schema",
+  "type": "object",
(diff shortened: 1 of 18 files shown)


ramls/dto/result.json
```diff
@@ -0,0 +1,37 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Result DTO Schema",
```
Igor-Gorchakov: better to rename the file caption from "result.json" to "resultDto.json", even if it's in "dto" directory

PR #28, opened 2019-02-25
MODSOURCE-27 Add sample records

KaterynaSenchenko:


mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SampleRecordsTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static org.folio.rest.impl.RecordApiTest.SOURCE_STORAGE_SOURCE_RECORDS_PATH;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+
+@RunWith(VertxUnitRunner.class)
+public class SampleRecordsTest extends AbstractRestVerticleTest {
+
+  @Override
+  public void clearTables(TestContext context) {
```
Igor-Gorchakov: Better to skip this ...

PR #32, opened 2019-03-11
Bug fix MODSOURCE-36 UUID in Sample Data should not be generated

TarasSpashchenko:
UUID for sample data must be the same as names of the files MARC records are contained in.


Igor-Gorchakov:
Please make Sonar happy

PR #33, opened 2019-03-14
MODSOURCE-33 Calculated generation numbers for new Records

KaterynaSenchenko:


mod-source-record-storage-server/src/test/java/org/folio/rest/impl/RecordsGenerationTest.java
```diff
@@ -0,0 +1,278 @@
...
+        });
+      });
+    });
+  }
+
+  @Test
+  public void shouldCalculateRecordsGeneration() {
+    List<Snapshot> snapshots = Arrays.asList(snapshot_1, snapshot_2, snapshot_3, snapshot_4);
+    for (int i = 0; i < snapshots.size(); i++) {
+      RestAssured.given()
```
OleksiiKuzminov: plz wrap each call into the Async

PR #40, opened 2019-04-03
MODSOURCE-38 - Add "Suppress Discovery" field to the SourceRecord data structure

RuslanLavrov:


mod-source-record-storage-server/src/test/java/org/folio/rest/impl/RecordApiTest.java
```diff
@@ -204,7 +207,8 @@ public void shouldReturnRecordsOnGetBySpecifiedSnapshotId(TestContext testContex
       .then()
       .statusCode(HttpStatus.SC_OK)
       .body("totalRecords", is(2))
-      .body("records*.snapshotId", everyItem(is(record_2.getSnapshotId())));
+      .body("records*.snapshotId", everyItem(is(record_2.getSnapshotId())))
+      .body("records*.additionalInfo.suppressDiscovery", everyItem(nullValue()));
```
KaterynaSenchenko: shouldn't it be false by default?

PR #43, opened 2019-04-16
MODSOURCE-16 - Change Delete method implemetation

RuslanLavrov:


mod-source-record-storage-server/src/test/java/org/folio/rest/impl/RecordApiTest.java
```diff
@@ -691,6 +830,7 @@ public void shouldCreateErrorRecordIfParsedContentIsInvalid(TestContext testCont
     Assert.assertThat(getRecord.getRawRecord().getContent(), is(rawRecord.getContent()));
     Assert.assertThat(getRecord.getParsedRecord(), nullValue());
     Assert.assertThat(getRecord.getErrorRecord(), notNullValue());
+    Assert.assertThat(getRecord.getDeleted(), is(false));
```
Igor-Gorchakov: .assertFalse

PR #44, opened 2019-04-18
MODSOURCE-47 Slow queries with 25k source records

Anatolii-Starkov:
-Hot fix\Hot Workaround

Here is the URL call

localhost:8081/source-storage/sourceRecords?query=Where source_records_view._id in  (SELECT records._id FROM records WHERE jsonb ->> 'recordType'  = 'MARC' LIMIT 50 OFFSET 200); &limit=0&offset=0


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -154,18 +158,33 @@
...
   public Future<SourceRecordCollection> getSourceRecords(String query, int offset, int limit, String tenantId) {
-    Future<Results<SourceRecord>> future = Future.future();
+
+    Future<ResultSet> future = Future.future();
     try {
-      String[] fieldList = {"*"};
-      CQLWrapper cql = getCQLWrapper(SOURCE_RECORDS_VIEW, query, limit, offset);
-      pgClientFactory.createInstance(tenantId).get(SOURCE_RECORDS_VIEW, SourceRecord.class, fieldList, cql, true, false, future.completer());
-    } catch (Exception e) {
+      pgClientFactory.createInstance(tenantId).select("SELECT * FROM source_records_view " + query, future.completer());
```
KaterynaSenchenko: Shouldn’t the query be parsed before attaching to the sql select? And what about limit and offset?

Anatolii-Starkov: Well, in perfect world yes it should be parsed, but it wasn't my aim here. I just want to share this approach with everybody, because my political and religious beliefs don't allow me to suggest it as work approach without an agreement of our team.  
What is about LIMIT and OFFSET. Did you have a chance to have a look at slack messages, or PR's description?
FYI
```localhost:8081/source-storage/sourceRecords?query=Where source_records_view._id in (SELECT records._id FROM records WHERE jsonb ->> 'recordType' = 'MARC' LIMIT 50 OFFSET 200); &limit=0&offset=0```

Comment (KaterynaSenchenko) [reply mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java]: ok, thanks, looks kinda strange though

File: mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SampleRecordsTest.java [hunk 2]
```diff
@@ -26,7 +26,7 @@ public void shouldReturnListWithSampleData() {
       .get(SOURCE_STORAGE_SOURCE_RECORDS_PATH)
       .then()
       .statusCode(HttpStatus.SC_OK)
-      .body("totalRecords", is(11))
+      .body("totalRecords", is(10))
```
KaterynaSenchenko: If I'm not mistaken, we have 11 sample records. Why did you change the expected number?

PR #46, opened 2019-04-24
Modsource 48

Anatolii-Starkov:
Added indexes for source record view in the records, raw_records, marc_records tables

### DIFF
--- a/mod-source-record-storage-server/src/main/resources/templates/db_scripts/create_records_view.sql
+++ b/mod-source-record-storage-server/src/main/resources/templates/db_scripts/create_records_view.sql
@@ -61,8 +61,8 @@ CREATE OR REPLACE VIEW source_records_view AS
      WHERE records.jsonb->>'parsedRecordId' IS NOT NULL;
 
 
-CREATE INDEX idx_btree_raw_record_id ON records USING BTREE ((jsonb ->> 'rawRecordId'));
-CREATE INDEX idx_btree_parsed_record_id ON records USING BTREE ((jsonb ->> 'parsedRecordId'));
-
-CREATE INDEX idx_btree_raw_records_id ON raw_records USING BTREE ((jsonb ->> 'id'));
-CREATE INDEX idx_btree_marc_records_id ON marc_records USING BTREE ((jsonb ->> 'id'));
+CREATE INDEX records_raw_record_id_idx_btree ON records USING BTREE ((jsonb ->> 'rawRecordId'));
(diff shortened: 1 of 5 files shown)


mod-source-record-storage-server/src/main/resources/templates/db_scripts/create_records_view.sql
```diff
@@ -59,3 +59,10 @@ CREATE OR REPLACE VIEW source_records_view AS
      JOIN raw_records ON records.jsonb->>'rawRecordId' = raw_records.jsonb->>'id'
      LEFT JOIN marc_records ON records.jsonb->>'parsedRecordId' = marc_records.jsonb->>'id'
      WHERE records.jsonb->>'parsedRecordId' IS NOT NULL;
+
+
+CREATE INDEX idx_btree_raw_record_id ON records USING BTREE ((jsonb ->> 'rawRecordId'));
+CREATE INDEX idx_btree_parsed_record_id ON records USING BTREE ((jsonb ->> 'parsedRecordId'));
+
+CREATE INDEX idx_btree_raw_records_id ON raw_records USING BTREE ((jsonb ->> 'id'));
```
OleksiiKuzminov: I think we need add for errorRecords index also

mod-source-record-storage-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -93,6 +93,18 @@
...
+      ],
+      "foreignKeys": [
+        {
+          "fieldName": "rawRecordId",
+          "targetTable": "raw_records",
+          "tOps": "ADD"
+        },
+        {
+          "fieldName": "parsedRecordId",
+          "targetTable": "marc_records",
```
KaterynaSenchenko: parsedRecord can be any type, not just MARC, therefore they will be stored in different tables. It is incorrect to put a foreignKey constraint on this field referencing marc_records table

Anatolii-Starkov: Please talk it over with @TarasSpashchenko. If it's bad approach the you will able to revert last commit and use Btree indexes implementation of that I had already pushed before.

Igor-Gorchakov: @Anatolii-Starkov , we've already discussed on Friday, using you changes made before.

KaterynaSenchenko: Reverted last commit. We will stick with btrees for now.

PR #47, opened 2019-04-26
MODSOURMAN-97 - Added fix to sourceRecord selection query building

RuslanLavrov:


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -101,9 +101,11 @@
...
     try {
       //Selects records from records table based on query and then retrieves data from source_records_view based on records.ids
-      CQLWrapper recordFilter = getCQLWrapper(RECORDS_TABLE, query, limit, offset);
-      recordFilter.addWrapper(getCQLWrapper(RECORDS_TABLE, "deleted=" + deletedRecords));
-      CQLWrapper countFilter = getCQLWrapper(RECORDS_TABLE, query);
+      CQLWrapper recordFilter = getCQLWrapper(RECORDS_TABLE, "deleted=" + deletedRecords, limit, offset);
+      recordFilter.addWrapper(getCQLWrapper(RECORDS_TABLE, query));
+
+      // trim sorting parameter from query for count filter to avoid invalid sql building
+      CQLWrapper countFilter = getCQLWrapper(RECORDS_TABLE, trimSortingParameter(query));
```
KaterynaSenchenko: I think we should add filtering of deleted here as well since we want to see the total of records satisfying the condition above.

PR #62, opened 2019-06-07
Modsource 55

Anatolii-Starkov:
-added errorMessages to the recordCollection DTO
-turned on auto generation of equals and hashCode in DTOs
-updated raml-storage
-fixed tests

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/RecordService.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/RecordService.java
@@ -1,12 +1,12 @@
 package org.folio.services;
 
+import java.util.Optional;
+
 import io.vertx.core.Future;
 import org.folio.rest.jaxrs.model.ParsedRecordCollection;
 import org.folio.rest.jaxrs.model.Record;
 import org.folio.rest.jaxrs.model.RecordCollection;
 import org.folio.rest.jaxrs.model.SourceRecordCollection;
-
-import java.util.Optional;
 
 /**
  * Record Service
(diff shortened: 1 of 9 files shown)


OleksiiKuzminov:
plz update news.md

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -69,11 +73,27 @@
...
+      .map(record -> Pair.of(record, saveRecord(record, tenantId)))
+      .collect(LinkedHashMap::new, (map, pair) -> map.put(pair.getKey(), pair.getValue()), Map::putAll);
+
+    Future<RecordCollection> result = Future.future();
+
+    CompositeFuture.join(new ArrayList<>(savedRecords.values())).setHandler(ar -> {
+        RecordCollection records = new RecordCollection();
+        savedRecords.forEach((record, future) -> {
+          if (future.failed()) {
+            records.getErrorMessages().add(future.cause().getMessage());
```
KaterynaSenchenko: consider adding some logging here

mod-source-record-storage-server/src/main/java/org/folio/services/RecordService.java
```diff
@@ -49,7 +49,7 @@
    * @param tenantId tenant id
    * @return future with true if succeeded
```
RuslanLavrov: please, update doc comment regarding return value

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
Anatolii-Starkov: Good point! I decided to add logging in rest layer in order to avoid flooding in a log

mod-source-record-storage-server/src/main/java/org/folio/services/RecordService.java
Anatolii-Starkov: Done

PR #64, opened 2019-06-12
MODSOURCE-56 Partial batch update for parsed records

Anatolii-Starkov:
-moved source-storsge/parsedRecordsCollection to a separate interface for batch operations
-added error messages to a response if any records were not updated

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,25 @@
+## 2019-06-12 v2.5.0
+* Changed url of updating parsed records from "/source-storage/parsedRecordsCollection" to "/source-storage/batch/parsed-records"
+* Added error message list to ParsedRecords DTO 
+
+ | METHOD |             URL                            | DESCRIPTION                                             |
+ |--------|--------------------------------------------|---------------------------------------------------------|
+ | GET    | /source-storage/snapshots                  | Get list of snapshots                                   |
+ | POST   | /source-storage/snapshots                  | Create new snapshot                                     |
+ | PUT    | /source-storage/snapshots/{jobExecutionId} | Update snapshot                                         |
+ | GET    | /source-storage/snapshots/{jobExecutionId} | Get snapshot by JobExecution id                         |
+ | DELETE | /source-storage/snapshots/{jobExecutionId} | Delete snapshot by JobExecution id                      |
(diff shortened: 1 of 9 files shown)


NEWS.md
```diff
@@ -3,6 +3,8 @@
 * Added endpoint to update a collection of ParsedRecords.
 * Added support for ParsedRecord.content represented in json.
 * Added endpoint for getting a formatted Record either by sourceRecordId or instanceId
+* Moved endpoint for saved record collection from /source-storage/recordsCollection to /source-storage/batch/records 
+* Changed response. List of error messages was added that is filled if some record was not saved
```
KaterynaSenchenko: we released the 2.4.0 version, so please start a new iteration in NEWS

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageBatchImpl.java
```diff
@@ -53,4 +54,26 @@ public void postSourceStorageBatchRecords(RecordCollection entity, Map<String, S
...
+  public void putSourceStorageBatchParsedRecords(ParsedRecordCollection entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    vertxContext.runOnContext(v -> {
+      try {
+        recordService.updateParsedRecords(entity, tenantId)
+                .map((ParsedRecordCollection it) -> {
+                  if (it.getErrorMessages().isEmpty()) {
+                    return (Response) PutSourceStorageBatchParsedRecordsResponse.respond200WithApplicationJson(it);
+                  } else {
+                    LOG.error("Some records were not saved! Here is the list of errors: {}", it.getErrorMessages());
+                    return (Response) PutSourceStorageBatchParsedRecordsResponse.respond500WithApplicationJson(it);
```
KaterynaSenchenko: I disagree with the 500 status here, but will back off if the team agrees that it's an appropriate solution. Let's move forward with the spike https://issues.folio.org/browse/MODSOURCE-63

OleksiiKuzminov: agree

PR #71, opened 2019-07-10
MODSOURCE-63

Anatolii-Starkov:
-draft

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
@@ -1,15 +1,23 @@
 package org.folio.services;
 
-import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
-import org.apache.commons.lang3.StringUtils;
 import org.folio.dao.RecordDao;
 import org.folio.dao.SnapshotDao;
-import org.folio.rest.jaxrs.model.*;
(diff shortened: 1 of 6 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -20,23 +8,29 @@
...
 import org.folio.dao.SnapshotDao;
-import org.folio.rest.jaxrs.model.AdditionalInfo;
-import org.folio.rest.jaxrs.model.ErrorRecord;
-import org.folio.rest.jaxrs.model.ParsedRecord;
-import org.folio.rest.jaxrs.model.ParsedRecordCollection;
-import org.folio.rest.jaxrs.model.Record;
-import org.folio.rest.jaxrs.model.RecordCollection;
-import org.folio.rest.jaxrs.model.SourceRecordCollection;
-import org.folio.rest.jaxrs.model.SourceStorageFormattedRecordsIdGetIdentifier;
+import org.folio.rest.jaxrs.model.*;
```
OleksiiKuzminov: imports

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -71,41 +111,36 @@
...
+      LOG.error("An error has occurred during saving a record: " + item.getErrorMessage());
+      future.complete(item);
+    }
+    return future;
+  }
 
-    Future<RecordCollection> result = Future.future();
+  private Item handleException(Item item, Throwable ex) {
+    if (ex instanceof NotFoundException) {
+      item.withStatus(404);
```
OleksiiKuzminov: plz use constants from the appropriate class 
"HttpStatus.SC_NOT_FOUND"

PR #77, opened 2019-08-08
Modinv 144

OleksiiKuzminov:
Overview: Libraries sometimes want to suppress records from discovery by their patrons. Much of the discovery work involves harvesting SRS MARC Bib records via OAI-PMH. When a user updates this flag in the FOLIO instance, the corresponding change should be made to the related SRS MARC Bib record, which then in turn updates the flag in the MARCcat MARC Bib record

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
@@ -180,9 +180,9 @@ public class RecordDaoImpl implements RecordDao {
     Future<Boolean> future = Future.future();
     String rollBackMessage = format("Record with %s id: %s was not found", suppressFromDiscoveryDto.getIncomingIdType().name(), suppressFromDiscoveryDto.getId());
     try {
-      String queryforRecordSearchByExternalId;
+      String queryForRecordSearchByExternalId;
       if (suppressFromDiscoveryDto.getIncomingIdType().equals(SuppressFromDiscoveryDto.IncomingIdType.INSTANCE)) {
-        queryforRecordSearchByExternalId = format(GET_RECORD_BY_INSTANCE_ID_QUERY, suppressFromDiscoveryDto.getId());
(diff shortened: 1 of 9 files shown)


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -170,6 +174,75 @@
...
+  public Future<Boolean> updateSuppressFromDiscoveryForRecord(SuppressFromDiscoveryDto suppressFromDiscoveryDto,
+                                                              String tenantId) {
+    Future<Boolean> future = Future.future();
+    String rollBackMessage = format("Record with %s id: %s was not found", suppressFromDiscoveryDto.getIncomingIdType().name(), suppressFromDiscoveryDto.getId());
+    try {
+      String queryforRecordSearchByExternalId;
+      if (suppressFromDiscoveryDto.getIncomingIdType().equals(SuppressFromDiscoveryDto.IncomingIdType.INSTANCE)) {
+        queryforRecordSearchByExternalId = format(GET_RECORD_BY_INSTANCE_ID_QUERY, suppressFromDiscoveryDto.getId());
+      } else {
+        throw new BadRequestException("Selected IncomingIdType doesn't supported");
```
Igor-Gorchakov:
```suggestion
        throw new BadRequestException("Selected IncomingIdType is not supported");
```

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -170,6 +174,75 @@
...
+        if (v.failed()) {
+          pgClientFactory.createInstance(tenantId).rollbackTx(tx, rollback -> future.fail(v.cause()));
+          return;
+        }
+        future.complete(true);
+      });
+      return future;
+    } catch (
+      Exception e) {
+      LOG.error("Error updating Record's suppress from discovery flag by {} id {}",
```
Igor-Gorchakov:
```suggestion
      LOG.error("Error while updating Record's suppress from discovery flag by {} id {}",
```

PR #78, opened 2019-08-09
MODSOURCE-70 Refactored DAOs to return saved/updated entities

KaterynaSenchenko:


KaterynaSenchenko:
Checked locally - looks OK, @OleksiiKuzminov, are you OK if I merge it?

PR #100, opened 2019-10-29
MODSOURCE-80 - Reduce the container Memory allocation in default LaunchDescriptor

RuslanLavrov:


dcrossleyau:
Please hold off with making any Memory configuration changes. We are investigating another option that might assist with this configuration.

PR #113, opened 2020-01-14
MODSOURCE-94 - Integrate mod-source-record-storage with mod-pubsub

RuslanLavrov:


mod-source-record-storage-server/pom.xml
```diff
@@ -86,6 +86,23 @@
       <version>1.0.0</version>
       <type>jar</type>
     </dependency>
+    <dependency>
+      <groupId>org.folio</groupId>
+      <artifactId>mod-pubsub-client</artifactId>
+      <version>1.1.0-SNAPSHOT</version>
```
KaterynaSenchenko: better use release version

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -58,22 +65,31 @@ public void postTenant(TenantAttributes entity, Map<String, String> headers, Han
...
         handlers.handle(ar);
       } else {
-        setLoadSampleParameter(entity, context)
-          .compose(v -> createStubSnapshot(context, entity))
-          .compose(v -> createStubData(entity))
-          .setHandler(event -> handlers.handle(ar));
+        registerModuleToPubsub(entity, headers, context.owner())
+          .setHandler(registrationAr -> {
+            if (registrationAr.failed()) {
+              handlers.handle(Future.failedFuture(registrationAr.cause()));
```
KaterynaSenchenko: I don't think we should fail tenant registration because of pubsub registration failure at this point. Just for now, let's not disrupt data-import flow when we transitioning towards pubsub.

PR #115, opened 2020-02-17
MODSOURCE-96 - Create event handler to update instance id to MARC bibliographic record

RuslanLavrov:


mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageHandlersImpl.java
```diff
@@ -17,21 +20,24 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(SourceStorageHandlersImpl.class);
 
+  @Autowired
+  EventHandlingService instanceCreatedEventHandleService;
```
OleksiiKuzminov: private?

mod-source-record-storage-server/src/main/java/org/folio/services/CreatedInstanceEventHandlingServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+
+import static org.folio.rest.jaxrs.model.EntityType.INSTANCE;
+import static org.folio.rest.jaxrs.model.EntityType.MARC_BIBLIOGRAPHIC;
+import static org.folio.services.util.AdditionalFieldsUtil.TAG_999;
+
+@Component()
```
OleksiiKuzminov: "()"

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageHandlersImpl.java
```diff
@@ -17,21 +20,24 @@
...
-      try {
-        // endpoint will be implemented in https://issues.folio.org/browse/MODSOURCE-96
-        LOG.info("Event was received: {}", entity);
-        Future.succeededFuture((Response) SourceStorageHandlers.PostSourceStorageHandlersCreatedInventoryInstanceResponse.respond200())
-          .setHandler(asyncResultHandler);
-      } catch (Exception e) {
-        LOG.error("Failed to handle event", e);
-        asyncResultHandler.handle(Future.succeededFuture(ExceptionHelper.mapExceptionToResponse(e)));
-      }
+      LOG.debug("Event CREATED_INVENTORY_INSTANCE was received: {}", entity);
```
VRohach: I think it will be better to log here only entity`s id. @RuslanLavrov @OleksiiKuzminov what do you think?

mod-source-record-storage-server/src/main/java/org/folio/services/CreatedInstanceEventHandlingServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Adds specified instanceId to record and additional custom field with instanceId to parsed record.
+   * Updates changed record in database.
+   *
+   * @param record      record to update
+   * @param instanceId  instance id
+   * @param tenantId    tenant id
+   * @return  future with updated parsed record
+   */
+  public Future<ParsedRecord> setInstanceIdToRecord(Record record, String instanceId, String tenantId) {
```
VRohach: Change to 'private'.

mod-source-record-storage-server/src/main/java/org/folio/services/CreatedInstanceEventHandlingServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+    if (record.getExternalIdsHolder() == null) {
```
VRohach: I think it will be better to use `Objects.isNull` here.

mod-source-record-storage-server/src/main/java/org/folio/services/CreatedInstanceEventHandlingServiceImpl.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public Future<Boolean> handle(String eventContent, String tenantId) {
+    try {
+      DataImportEventPayload dataImportEventPayload = ObjectMapperTool.getMapper().readValue(eventContent, DataImportEventPayload.class);
+      String instanceAsString = dataImportEventPayload.getContext().get(INSTANCE.value());
+      String recordAsString = dataImportEventPayload.getContext().get(MARC_BIBLIOGRAPHIC.value());
+
+      if (StringUtils.isEmpty(instanceAsString) || StringUtils.isEmpty(recordAsString)) {
+        String msg = "Failed to handle CREATED_INVENTORY_INSTANCE event, cause event payload context does not contain INSTANCE and MARC_BIBLIOGRAPHIC data";
```
VRohach: I think it will be better to extract this variable to the constant.

VRohach: Please replace "does not contain INSTANCE and MARC_BIBLIOGRAPHIC data" on "does not contain INSTANCE and/or MARC_BIBLIOGRAPHIC data".

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Adds field if it does not exist and a subfield with a value to that field
+   *
+   * @param record   record that needs to be updated
+   * @param field    field that should contain new subfield
+   * @param subfield new subfield to add
+   * @param value    value of the subfield to add
+   * @return true if succeeded, false otherwise
+   */
+  public static boolean addFieldToMarcRecord(Record record, String field, char subfield, String value) {
```
VRohach: Change to 'void' or use 'boolean' return value. Because for current implementation return value never using.

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -0,0 +1,92 @@
...
+    boolean result = false;
+    try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
```
VRohach: Can we use here try-with-resources?

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -0,0 +1,92 @@
...
+        MarcJsonWriter jsonWriter = new MarcJsonWriter(os);
+        MarcFactory factory = MarcFactory.newInstance();
+        if (reader.hasNext()) {
+          org.marc4j.marc.Record marcRecord = reader.next();
+          VariableField variableField = getSingleFieldByIndicators(marcRecord.getVariableFields(field), INDICATOR, INDICATOR);
+          DataField dataField;
+          if (variableField != null
+            && ((DataField) variableField).getIndicator1() == INDICATOR
+            && ((DataField) variableField).getIndicator2() == INDICATOR
+          ) {
```
VRohach: Move `)` symbol

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -0,0 +1,92 @@
...
+      LOGGER.error("Failed to add additional subfield {} for field {} to record {}", e, subfield, field, record.getId());
+    }
+    return result;
+  }
+
+  private static MarcReader buildMarcReader(Record record) {
+    return new MarcJsonReader(new ByteArrayInputStream(record.getParsedRecord().getContent().toString().getBytes(StandardCharsets.UTF_8)));
+  }
+
+  private static VariableField getSingleFieldByIndicators(List<VariableField> list, char ind1, char ind2) {
```
VRohach: @RuslanLavrov why did you pass there these arguments: "ind1", "ind2"? Regarding current logic, they are always  "f".

mod-source-record-storage-server/src/main/java/org/folio/services/CreatedInstanceEventHandlingServiceImpl.java
RuslanLavrov: In my opinion, Objects.isNull has not essential advantages for this situation, also we avoid redundant method call.

VRohach: @RuslanLavrov Yes, you are right.

mod-source-record-storage-server/src/main/java/org/folio/services/CreatedInstanceEventHandlingServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+
+import static org.folio.rest.jaxrs.model.EntityType.INSTANCE;
+import static org.folio.rest.jaxrs.model.EntityType.MARC_BIBLIOGRAPHIC;
+import static org.folio.services.util.AdditionalFieldsUtil.TAG_999;
+
+@Component
+public class CreatedInstanceEventHandlingServiceImpl implements EventHandlingService {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CreatedInstanceEventHandlingServiceImpl.class);
+  private static final String EVENT_HAS_NOT_NEEDED_DATA_MSG = "Failed to handle CREATED_INVENTORY_INSTANCE event, cause event payload context does not contain INSTANCE and/or MARC_BIBLIOGRAPHIC data";
```
KaterynaSenchenko:
```suggestion
  private static final String EVENT_HAS_NO_DATA_MSG = "Failed to handle CREATED_INVENTORY_INSTANCE event, because event payload context does not contain INSTANCE and/or MARC_BIBLIOGRAPHIC data";
```

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+    return result;
+  }
+
+  private static MarcReader buildMarcReader(Record record) {
+    return new MarcJsonReader(new ByteArrayInputStream(record.getParsedRecord().getContent().toString().getBytes(StandardCharsets.UTF_8)));
+  }
+
+  private static VariableField getSingleFieldByIndicators(List<VariableField> list) {
+    if (list == null || list.isEmpty()) {
```
KaterynaSenchenko: CollectionUtils.isEmpty(list) ?

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageHandlersImpl.java
OleksiiKuzminov: It's debug and we don't have an id

VRohach: @OleksiiKuzminov thx for your answer.

PR #122, opened 2020-03-20
MODSOURCE-111 Fixed searching SourceRecords by recordId

KaterynaSenchenko:


grabsefx:
It's a good practice to name pull requests starting the jira name. Something like:
"MODSOURCE-111 Pretty nice sourceRecords fix". In this case current PR could be linked to jira automatically.
Thanks for the fix, anyway )

KaterynaSenchenko:
Hi @grabsefx, thanks. I didn't link this PR to the jira ticket hoping to find more elegant solution. But it looks like we'll go with this one for now.

PR #128, opened 2020-04-03
MODSOURCE-118: Archive event payload.

VRohach:
1. Unzipping mechanism added.
2. "data-import-processing-core" was added to the pom.xml.


mod-source-record-storage-server/src/test/java/org/folio/services/CreatedInstanceEventHandlingServiceImplTest.java
```diff
@@ -110,7 +111,14 @@ public void shouldSetInstanceIdToRecord(TestContext context) {
       .withContext(payloadContext);
 
     Future<Boolean> future = recordDao.saveRecord(record, TENANT_ID)
-      .compose(rec -> eventHandlingService.handle(Json.encode(dataImportEventPayload), TENANT_ID));
+      .compose(rec -> {
+        try {
+          return eventHandlingService.handle(ZIPArchiver.zip(Json.encode(dataImportEventPayload)), TENANT_ID);
+        } catch (IOException e) {
+          e.printStackTrace();
+          return null;
```
KaterynaSenchenko, line 119: return failed future? this way we should have an assertion error in the handler

VRohach: Changed.

PR #131, opened 2020-04-07
MODSOURCE-114: Add State "field" to records table.

VRohach:
1.Mechanism for retrieving via "matchedId" and "state" was added.
2. Index on "matchedId" and "state" added.
3. State "ACTUAL" was added to the test data.
4. Tests added.


VRohach:
Other tests will be fixed and added soon.

mod-source-record-storage-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -82,7 +82,7 @@
...
       "fromModuleVersion": "mod-source-record-storage-3.1.0"
```
RuslanLavrov, line 86: please update here "fromModuleVersion" to mod-source-record-storage-3.2.0

OleksiiKuzminov:
plz add changes to test data and add state field and ok

PR #133, opened 2020-04-15
MODOAIPMH-119 - Extend SourceRecord dto with externalIdsHolder field.

RuslanLavrov:


NEWS.md
```diff
@@ -1,6 +1,9 @@
 ## 2020-04-10 v3.2.0-SNAPSHOT
 * Added State "field" to records table. (MODSOURCE-114)
 
+## 2020-04-07 v3.1.4-SNAPSHOT
```
KaterynaSenchenko, line 4: do we need it as a hotfix on fameflower?

RuslanLavrov: I discussed this with @OleksiiKuzminov, he confirmed that we need

PR #135, opened 2020-04-16
MODSOURCE-115: Create endpoint for retrieving "sourceRecords" using "matchedId" and generationId-mechanism.

VRohach:
1. New endpoint for retrieving "source  records" was added.
2. Signature: /source-storage/sourceRecords/\{id\}?idType=INSTANCE
3. New function "get_source_record_by_external_id(externalId uuid, idFieldName text)" for DB was added. (It is based on searching sourceRecord via specific externalIdsHolder for specific external id type).
4. Existing DB function was changed on retrieving "sourceRecords" via "matchedId" and "ACTUAL" state. Moreover, was added "externalIdsHolders"-field to the resulted jsonb of these functions.
5. The mechanism for extending future external id type - logic was added.
6. Tests created.


VRohach:
Code coverage will be increased soon.

mod-source-record-storage-client/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.folio</groupId>
     <artifactId>mod-source-record-storage</artifactId>
-    <version>3.2.0-SNAPSHOT</version>
+    <version>3.2.1-SNAPSHOT</version>
```
KaterynaSenchenko, line 9: what is the reason for bumping the development version? new endpoint is a backward compatible change, it is definitely not a bugfix, and I strongly oppose incrementing snapshot versions in any case (unless it is absolutely necessary, which is not the case here)

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -175,6 +177,32 @@
     });
   }
 
+  @Override
+  public Future<Optional<SourceRecord>> getSourceRecord(String id, ExternalIdType externalIdType, String tenantId) {
```
KaterynaSenchenko, line 181: wouldn't it be more convenient to have separate dao methods to retrieve sourceRecord by id (aka matchedId) and by externalId of any kind?

VRohach: Sure, will do it.

NEWS.md
```diff
@@ -1,3 +1,6 @@
+## v3.2.1-SNAPSHOT
+* [MODSOURCE-115](https://issues.folio.org/browse/MODSOURCE-115) Create endpoint for retrieving "sourceRecords" using "matchedId" and generationId-mechanism.
```
RuslanLavrov, line 2: please move it to v3.1.4-SNAPSHOT section if we really need to release these changes as hotfix, otherwise to v3.2.0-SNAPSHOT section

KaterynaSenchenko: nope, this is not a bugfix, current development iteration is 3.2.0-SNAPSHOT

NEWS.md
```diff
@@ -1,3 +1,6 @@
+## v3.2.1-SNAPSHOT
```
KaterynaSenchenko, line 1:
```suggestion
## v3.2.0-SNAPSHOT
```
!!!!!

descriptors/ModuleDescriptor-template.json
```diff
@@ -297,6 +306,11 @@
       "displayName": "Source Storage - get results",
       "description": "Get Results"
     },
+    {
+      "permissionName": "source-storage.sourceRecords.get",
```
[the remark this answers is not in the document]
VRohach: @KaterynaSenchenko I added completely new permission, "sourcreRecord" (with single number) for retreiving a single source record. Do you think that this is not really good idea?

mod-source-record-storage-client/pom.xml
VRohach: @KaterynaSenchenko Sure. Thank you for you explanation!

NEWS.md
VRohach: Yes, I changed it in my last commit

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
KaterynaSenchenko: it this case - see suggested change, "sourceRecord" name in permission should be singular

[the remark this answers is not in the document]
VRohach: Done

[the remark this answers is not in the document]
VRohach: @KaterynaSenchenko @RuslanLavrov  So I changed. Now, I`m using an existing one, as you recommended.

PR #148, opened 2020-04-24
MODSOURCE-116: Add filtering of sourceRecords to return only latest generation.

VRohach:
1. The mechanism for putting records with a specific state created.
2. Condition on "state"=ACTUAL for '/source-storage/records' was added (scripts and DAO changed).
3. Condition on "state"=ACTUAL for '/source-storage/sourceRecords' was added (script changed).
4. Tests changed for covered different cases with the state field.


VRohach:
Guys, I want to mention, that in the scope of this PR was added mechanism for retrieving records with "ACTUAL" state for endpoint: '/source-storage/records'. 

Moreover, was added a mechanism for putting record with different from "ACTUAL". (It is needed for testing purposes. Before that, ACTUAL state was ALWAYS, even in those cases when we put record with another state.)

PR #149, opened 2020-04-28
MODSOURCE-130: Liquibase standard table creation

wwelling:
[MODSOURCE-130](https://issues.folio.org/browse/MODSOURCE-130)

Based off [PrepareSRSSchema.v1.5.sql](https://issues.folio.org/secure/attachment/26516/PrepareSRSSchema.v1.5.sql) from umbrella issue [MODSOURCE-97](https://issues.folio.org/browse/MODSOURCE-97)

Here is a plain database dump of `diku_mod_source_record_storage` schema after Liquibase migration for tenant diku:
[mod-source-record-storage.sql.txt](https://github.com/folio-org/mod-source-record-storage/files/4546145/mod-source-record-storage.sql.txt)


mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-0.0.1/2020-04-27--12-00-create-snapshots-table.xml
```diff
@@ -0,0 +1,39 @@
...
+  <changeSet id="2020-04-27--12-01-create-snapshots-table" author="WilliamWelling">
+    <createTable tableName="snapshots_lb">
+      <column name="id" type="uuid">
+        <constraints primaryKey="true" nullable="false"/>
+      </column>
+      <column name="status" type="${database.defaultSchemaName}.job_execution_status">
+        <constraints nullable="false"/>
+      </column>
+      <column name="processing_started_date" type="timestamp">
+        <constraints nullable="false"/>
```
KaterynaSenchenko, line 34: I believe at the point of creation of a new Snapshot (with status NEW) processing_started_date can be null

wwelling: Updated. Thanks.

> I believe at the point of creation of a new Snapshot (with status NEW) processing_started_date can be null

Removed null constraint. Thank you.

PR #150, opened 2020-05-04
MODSOURCE-102: standard table data access objects

wwelling:
Blocked by https://github.com/folio-org/data-import-raml-storage/pull/169

[MODSOURCE-102](https://issues.folio.org/browse/MODSOURCE-102)


OleksiiKuzminov:
@wwelling  In general, it looks awsome
I need more time to check all changes, but this approach is something like I planed

mod-source-record-storage-server/src/main/java/org/folio/dao/impl/LBRecordDaoImpl.java
```diff
@@ -0,0 +1,252 @@
...
+      params.addNull().addNull().addNull().addNull();
+    }
+    return params;
+  }
+
+  @Override
+  protected RecordCollection toCollection(ResultSet resultSet) {
+    return new RecordCollection()
+      .withRecords(resultSet.getRows().stream().map(this::toEntity).collect(Collectors.toList()))
+      .withTotalRecords(resultSet.getNumRows());
```
RuslanLavrov, line 177: @OleksiiKuzminov @TarasSpashchenko do we need to return a number of all records which conform filter condition in the "totalRecords" field to support pagination? As I understand resultSet.getRows().size() == resultSet.getNumRows(), and resultSet.getNumRows() depends not only on filter condition but on limit and offset as well.

wwelling: Correct. Good observation. Looks like we will need a count query.

wwelling: Would it be appropriate to not set this here and perform the count query at the service layer?

wwelling: Maybe it would be more straight forward to augment the queries with count response. I will update accordingly.

RuslanLavrov: Thanks William!

wwelling: Will this solution work for now? Should be performant enough with Postgres 9.2 improvements to count query. The only caveat is if no rows are returned the total is unknown, assigned -1 for now. This can be remedied by more advanced query or splitting count query from actual query.

RuslanLavrov: I think it will be better to improve it if it's possible.

wwelling: I have updated the queries to always return total count. If there are no actual result rows, there will be just a single row with the total count present. This can occur if limit is 0 or if offset is greater than the limit. I have updated logic to handle this result set and added tests accordingly.

RuslanLavrov: Thank you for the improvement!

PR #153, opened 2020-05-07
MODSOURCE-102: upgrade to vertx-pg-client

wwelling:
This PR will change to pull into master after https://github.com/folio-org/mod-source-record-storage/pull/150 for better change comparison.

I would like to propose this in anticipation to the [RMB](https://github.com/folio-org/raml-module-builder/pull/640) switch to [vertx-pg-client](https://vertx.io/docs/vertx-pg-client/java/). Most changes will be required. Changes include value template character from `?` to `$n`, SQL template parameters from `JsonArray` to `Tuple`, and async database response `ResultSet` to `RowSet<Row>`. Direct use of the vertx-pg-client affords batch, cursor, transactions, streaming, etc. without limitations of intermediate class. The tenant pool map in PostgresClientFactory can either evolve into moving away from RMB or simply switch to use PgPool client exposed from RMB PostgresClient, if it becomes public accessible.


wwelling:
https://github.com/folio-org/raml-module-builder/pull/640/commits/404c2ad5c0e45e7019b67ae2b9f74f3adf9729b9

We now have a clear path using PostgresClient. When upgrading RMB, we can remove the manual PgPool creation in PostgresClientFactory.

wwelling:
Closing to reopen against master branch.

PR #154, opened 2020-05-12
MODINV-207 Implemented handler for QM update

KaterynaSenchenko:


mod-source-record-storage-server/src/main/java/org/folio/services/UpdateRecordEventHandler.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Handles QM_RECORD_UPDATED event and sends QM_SRS_MARC_BIB_RECORD_UPDATED as a result of the update
+   *
+   * @param eventContent event content
+   * @param params       OkapiConnectionParams
+   * @return future with true if succeeded
+   */
+  public Future<Boolean> handle(String eventContent, OkapiConnectionParams params) {
+    try {
+      HashMap<String, String> eventPayload = ObjectMapperTool.getMapper().readValue(ZIPArchiver.unzip(eventContent), HashMap.class);
```
VRohach, line 40: I think it will be better to handle here exception from unzip()-method with the error message "Can`t unzip eventContent".
Moreover, let`s change IOException in 56 line to the common Exception in case of other errors

PR #156, opened 2020-05-13
MODSOURCE-102.1: upgrade to vertx-pg-client

wwelling:
I would like to propose this in anticipation to the [RMB](https://github.com/folio-org/raml-module-builder/pull/640) switch to [vertx-pg-client](https://vertx.io/docs/vertx-pg-client/java/). Most changes will be required. Changes include value template character from `?` to `$n`, SQL template parameters from `JsonArray` to `Tuple`, and async database response `ResultSet` to `RowSet<Row>`. Direct use of the vertx-pg-client affords batch, cursor, transactions, streaming, etc. without limitations of intermediate class. The tenant pool map in `PostgresClientFactory` can either evolve into moving away from RMB or simply switch to use `PgConnection` which will be exposed from RMB `PostgresClient`, if it becomes public accessible.

The commits represent changes of MODSOURCE-102 branch as well. The previous PR was squashed  and merged.


mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
```diff
@@ -1,13 +1,33 @@
...
+  private static final String PORT = "port";
+  private static final String DATABASE = "database";
+  private static final String PASSWORD = "password";
+  private static final String USERNAME = "username";
+
+  private static final String DEFAULT_SCHEMA_PROPERTY = "search_path";
+
+  private static final int POOL_SIZE = 5;
+
+  private static final Map<String, PgPool> pool = new HashMap<>();
```
OleksiiKuzminov, line 29: Can it be cunccurentHashMap and "private synchronized PgPool getPool(String tenantId)" make static without synchronized?

OleksiiKuzminov: or just method without synchronized

mod-source-record-storage-server/src/main/java/org/folio/dao/AbstractEntityDao.java
```diff
@@ -61,109 +64,126 @@ public void getByQuery(Q query, int offset, int limit, String tenantId, Handler<
...
+      SqlConnection connection = ar1.result();
+      connection.prepare(sql, ar2 -> {
+        if (ar2.failed()) {
+          log.error("Failed to prepare query", ar2.cause());
+          endHandler.handle(Future.failedFuture(ar2.cause()));
           return;
         }
-        stream.result()
+        PreparedStatement pq = ar2.result();
+        Transaction tx = connection.begin();
```
OleksiiKuzminov, line 82: Can we move transaction into the DaoUtil and create a template method?

mod-source-record-storage-server/src/main/java/org/folio/dao/impl/SourceRecordDaoImpl.java
```diff
@@ -158,107 +166,127 @@ public void getSourceMarcRecordsByQuery(SourceRecordContent content, RecordQuery
...
+        Transaction tx = connection.begin();
+        RowStream<Row> stream = pq.createStream(limit, Tuple.tuple());
+        stream
           .handler(row -> {
-            stream.result().pause();
-            lookupContent(content, tenantId, toSourceRecord(row)).setHandler(res -> {
-              if (res.failed()) {
-                endHandler.handle(Future.failedFuture(res.cause()));
+            stream.pause();
+            lookupContent(content, tenantId, toSourceRecord(row)).onComplete(ar3 -> {
```
OleksiiKuzminov, line 189: Should it be moved to the Service layer

mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
wwelling: If map is concurrent and method is not synchronized, it would be possible to have two `PgPool` being created at the same time and the one that completes second would replace the first put in the map. We currently cannot make the method static as still require the vertx injected bean. I think an issue is justified to create `PgPool` without `PostgresClient` to retrieve database connection config. Then we can simplify and control the map better.

OleksiiKuzminov: ok, my main concern that synchronized methods can slowdown vertx event loop. So we need to think about the async approach or figure out way to deal with it after RMB updates. Can you create a task to cover this issue?

mod-source-record-storage-server/src/main/java/org/folio/dao/AbstractEntityDao.java
wwelling: Done

mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
wwelling: Yes, I will create the issue for asynchronous access to get client. Should there be another for getting connection configuration without PostgresClient?

wwelling: Done. [MODSOURCE-138](https://issues.folio.org/projects/MODSOURCE/issues/MODSOURCE-138)

PR #159, opened 2020-05-20
MODSOURCE-137: port existing service logic to standard table DAO services

wwelling:
[MODSOURCE-137](https://issues.folio.org/browse/MODSOURCE-137)


mod-source-record-storage-server/src/main/java/org/folio/dao/query/RecordQuery.java
```diff
@@ -11,41 +10,37 @@
...
-
-  private RecordQuery() {
-    super(ImmutableMap.copyOf(ptc), Record.class);
+@Metamodel(
+  entity = Record.class,
+  table = RECORDS_TABLE_NAME,
+  properties = {
+    @Property(path = "id", column = ID_COLUMN_NAME),
+    @Property(path = "snapshotId", column = SNAPSHOT_ID_COLUMN_NAME),
+    @Property(path = "matchedProfileId", column = MATCHED_PROFILE_ID_COLUMN_NAME),
```
KaterynaSenchenko, line 29: @OleksiiKuzminov, is that a deprecated property? I don't think we use it anywhere. If so, maybe we should get rid of at this point.

OleksiiKuzminov: yep "matchedProfileId" is not used anywhere

mod-source-record-storage-server/src/main/java/org/folio/services/impl/LBRecordServiceImpl.java
```diff
@@ -1,15 +1,259 @@
...
+
+  @Override
+  public Future<Record> updateSourceRecord(ParsedRecordDto parsedRecordDto, String snapshotId, String tenantId) {
+    String id = parsedRecordDto.getId();
+    // no processing of the record is performed apart from the update itself
+    Snapshot snapshot = new Snapshot()
+      .withJobExecutionId(snapshotId)
+      .withStatus(Snapshot.Status.COMMITTED);
+    return dao.inTransaction(tenantId, connection ->
+      rawRecordDao.getById(connection, id, tenantId)
```
KaterynaSenchenko, line 160: I'm not sure I understand why we need to search for rawRecord if it is not updated by this operation. And as for the id value of the record, would it match rawRecord id?

wwelling: The new record table id is now a foreign key for raw record table id, marc record table id, and error record table id. This look is performed to get the existing raw record to add it to the new record created. Previously, the record lookup via SQL function with join to all related tables returned the raw record. The save of new record and old record follows previous save logic.

wwelling: There is absolutely room for optimization here, but did not feel it justified as it is only updating single source record.

wwelling: Optimization would require changes in [saveAndUpdate](https://github.com/folio-org/mod-source-record-storage/pull/159/files/22c9fa7e613a93cf02a23086a75bf8e41c786f16#diff-08b1dfa84e6c6cae8956502ce38a64c8R232), [insertOrUpdateRecord](https://github.com/folio-org/mod-source-record-storage/pull/159/files/22c9fa7e613a93cf02a23086a75bf8e41c786f16#diff-08b1dfa84e6c6cae8956502ce38a64c8R195), etc.

mod-source-record-storage-server/src/main/java/org/folio/dao/query/RecordQuery.java
wwelling: Would you like it removed in this PR? I do not think the JSON schema change should occur yet though.

KaterynaSenchenko: I'll create a small task for update of the json schema - it should be addressed separately as there are dependencies in other modules for the record.json, those modules should be updated as well. 
If it is safe to delete this property here without the schema update, please go ahead

wwelling: Thanks. Will remove from new records table and any reference in the new DAOs and Services.

mod-source-record-storage-server/src/main/java/org/folio/services/impl/LBRecordServiceImpl.java
KaterynaSenchenko: OK, got it, thanks

mod-source-record-storage-server/src/main/java/org/folio/dao/query/RecordQuery.java
KaterynaSenchenko: Thanks, and here is the task for updating the schema https://issues.folio.org/browse/MODSOURCE-145

PR #160, opened 2020-05-26
MODSOURCE-134: Add new RAML and update module descriptor

wwelling:
[MODSOURCE-134](https://issues.folio.org/browse/MODSOURCE-134)

New Source Record Storage RAML replacing CQL with query parameters for existing requests.


OleksiiKuzminov:
Hello @wwelling. We need to update a module descriptor and increment interface version
Also what about schema changes? Do you think current schemas and DTO's are ok?

wwelling:
@OleksiiKuzminov I had forgot all about the module descriptor. Let me get that updated. As for the schemas, I do not think any need to be changed for the endpoints for the new raml. Some may need to be removed during cleanup of old endpoints. If we decide to go with the jOOQ implementation of DAOs, I would recommend we update the schemas to match the jOOQ generated POJOs to simplify mapping to and from the database.

wwelling:
@OleksiiKuzminov updated module descriptor

OleksiiKuzminov:
I believe we can delete old raml and rename endpoints without lb prefix

wwelling:
I did not want to make any breaking changes for this issue. I believe we plan for old service/api cleanup in issue MODSOURCE-144. It will be trivial to refactor raml and module descriptor when new endpoints are implement. Would it be ok to postpone the removal of old raml and refactoring until then?

PR #178, opened 2020-06-13
MODSOURCE-144: migration from old tables to new

wwelling:
- reverts changes to existing liquibase scripts on 3.2.0
- drop tables and types from liquibase scripts on 3.2.0
- create new refactored liquibase tables
- migrate from old JSONB tables to new standard tables
- drop old JSONB tables
- drop unused functions


RuslanLavrov:
Hello @wellingWilliam, could you please remove the scripts related to the creation of the non-tenant schema "source_record_storage_config". Looks like this schema will not be used.

https://github.com/folio-org/mod-source-record-storage/tree/master/mod-source-record-storage-server/src/main/resources/liquibase/module
https://github.com/folio-org/mod-source-record-storage/blob/e2efc62662d23fea519ec9049e6f4127944b9c63/mod-source-record-storage-server/src/main/java/org/folio/rest/impl/InitAPIImpl.java#L30

PR #187, opened 2020-06-17
MODSOURCE-156: Get source records by list of ids

wwelling:
https://issues.folio.org/browse/MODSOURCE-156

This is fast forward from #186 to avoid conflicts.


KaterynaSenchenko:
@wwelling, could you please change the version to 4.0.0-SNAPSHOT? It should make it easier to test the updates in other modules and avoid any confusion

wwelling:
Yeah. I am also testing migration before merging.

PR #195, opened 2020-06-22
Add deleted query parameter and appropriate conditions

wwelling:


mod-source-record-storage-server/src/main/java/org/folio/dao/util/RecordDaoUtil.java
```diff
@@ -422,6 +426,23 @@ public static Condition filterRecordByUpdatedDateRange(Date updatedAfter, Date u
...
 
+  /**
+   * Get {@link Condition} to filter by state ACTUAL or DELETED or leader record status d, s, or x
+   * 
+   * @param deleted deleted flag
+   * @return condition
+   */
+  public static Condition filterRecordByDeleted(Boolean deleted) {
+    Condition condition = filterRecordByState(RecordState.ACTUAL.name());
+    if (Objects.isNull(deleted) || Boolean.TRUE.equals(deleted)) {
```
KaterynaSenchenko, line 437: include deleted when 'deleted' param is null? or is it set to false by default anyway?

wwelling: It is default false in RAML and should never be null. I will update this condition. Thanks.

KaterynaSenchenko: Great! Thanks

PR #198, opened 2020-06-23
Deleted db schema definition from schema.json

RuslanLavrov:
db schema definition has been deleted from the schema.json because Liquibase is used as db migration tool starting with v4.0.0 module version


wwelling:
If this works as expected, we can update the Liquibase migration scripts that remove these tables. They are currently set to always run. `runAlways="true"` This can be removed as the tables will be gone for good now.

https://github.com/folio-org/mod-source-record-storage/blob/refactoring-master-erase-schema-json/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-0.0.2/2020-06-13--12-00-drop-error-records.xml#L7
https://github.com/folio-org/mod-source-record-storage/blob/refactoring-master-erase-schema-json/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-0.0.2/2020-06-13--13-00-drop-marc-records.xml#L7
https://github.com/folio-org/mod-source-record-storage/blob/refactoring-master-erase-schema-json/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-0.0.2/2020-06-13--14-00-drop-raw-records.xml#L7
https://github.com/folio-org/mod-source-record-storage/blob/refactoring-master-erase-schema-json/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-0.0.2/2020-06-13--15-00-drop-records.xml#L7
https://github.com/folio-org/mod-source-record-storage/blob/refactoring-master-erase-schema-json/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-0.0.2/2020-06-13--16-00-drop-snapshots.xml#L7
https://github.com/folio-org/mod-source-record-storage/blob/refactoring-master-erase-schema-json/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-0.0.2/2020-06-13--17-00-drop-unused-functions.xml

RuslanLavrov:
@wwelling, I've updated scripts. Thank you

PR #202, opened 2020-06-24
Fix date format

mrdpopov:
# Purpose

Fixes date formats in SRS client


KaterynaSenchenko:
Just beware that after RMB upgrade, client will be RMB generated again. Will need to check when we remove this classes

mrdpopov:
Will it have the correct date format?

KaterynaSenchenko:
it will, it was a note for Folijet to pay attention when switching back to generated client

PR #203, opened 2020-06-24
MODSOURCE-164: Use vert.x bom for vertx and jackson

julianladisch:
No longer downgrade jackson from 2.10.2 to 2.10.0.pr1.
Use vertx-stack-depchain bill of materials (bom) for vert.x and
jackson dependencies.


julianladisch:
The Jira for this pull request is https://issues.folio.org/browse/MODSOURCE-164
Core platform team does not have write permissions for mod-source-record-storage repository. Therefore Folijet team needs to also merge the pull request after it has been approved in code review

PR #207, opened 2020-07-01
MODSOURCE-163: Use UTC instead of system default

wwelling:
This should resolve [MODSOURCE-163](https://issues.folio.org/browse/MODSOURCE-163)

Here is the [PR for RMB](https://github.com/folio-org/raml-module-builder/pull/721) for next release to replace clients with generated clients.


KaterynaSenchenko:
Merging this PR in order to include it in 4.0.1 bugfix release

PR #230, opened 2020-09-16
MODDICORE-58: MARC-MARC matching for 001 and 999 ff fields

VRohach:
1. New handler for matching MARC_BIB records created and registered.
2. Messaging descriptor changed: added new events.
3. Putting "instanceHrid" to the DB added.


VRohach:
FYI: tests, news, and docs for the new logic will be written in the nearest future.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/MarcBibliographicMatchEventHandler.java
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * Handles MARC_BIB_CREATED event.
+   */
+  @Override
+  public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload dataImportEventPayload) {
+    CompletableFuture<DataImportEventPayload> future = new CompletableFuture<>();
+    HashMap<String, String> context = dataImportEventPayload.getContext();
+    dataImportEventPayload.getEventsChain().add(dataImportEventPayload.getEventType());
+    String recordAsString = context.get(MARC_BIBLIOGRAPHIC.value());
```
KaterynaSenchenko, line 75: please add checks for null and look for other places where it might be applicable

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/MarcBibliographicMatchEventHandler.java
```diff
@@ -0,0 +1,200 @@
...
+    if (StringUtils.isEmpty(recordAsString)) {
+      LOG.error(EVENT_HAS_NO_DATA_MSG);
+      future.completeExceptionally(new EventProcessingException(EVENT_HAS_NO_DATA_MSG));
+      return future;
+    }
+
+    MatchDetail matchDetail = retrieveMatchDetail(dataImportEventPayload);
+    String valueFromField = retrieveValueFromMarcFile(recordAsString, matchDetail.getIncomingMatchExpression());
```
KaterynaSenchenko, line 83:
```suggestion
    String valueFromField = retrieveValueFromMarcRecord(recordAsString, matchDetail.getIncomingMatchExpression());
```

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/MarcBibliographicMatchEventHandler.java
VRohach: @KaterynaSenchenko sure! I added all additional check for eventPayload data.

RuslanLavrov (approved):

mod-source-record-storage-server/src/test/java/org/folio/services/MarcBibliographicMatchEventHandlerTest.java
```diff
@@ -0,0 +1,412 @@
...
+            .withIncomingMatchExpression(new MatchExpression()
+              .withDataValueType(VALUE_FROM_RECORD)
+              .withFields(Lists.newArrayList(
+                new Field().withLabel("field").withValue("948"),
+                new Field().withLabel("indicator1").withValue(""),
+                new Field().withLabel("indicator2").withValue(""),
+                new Field().withLabel("recordSubfield").withValue("a"))))))));
+
+    CompletableFuture<DataImportEventPayload> future = new CompletableFuture<>();
+    recordDao.saveRecord(record, TENANT_ID)
```
RuslanLavrov, line 195: I propose to move call of the method recordDao.saveRecord() to the setUp() method to simplify async coordination code and avoid hanging of test in case saveRecord() returns failed future

KaterynaSenchenko (approved):

PR #267, opened 2021-01-08
MODSOURCE-227 PoC

Igor-Gorchakov:
# Purpose
The purpose of PoC is to check how to enable Streaming API in the SRS using the standard RMB feature.
We do not merge this right now. This PR will hang unmerged with a purpose to apply the changes later.

# Approach
1. Configure both pom.xml files to generate bulk endpoint with RoutingContext;
The root pom.xml is used to turn on generating bulk endpoint,
The server pom.xml is used to initialize the RountingContext;

2. Use RoutingContext to get an HttpServiceResponse;

3. Use WriteStream.pipeTo to take data from the RowStream(implements ReadStream) and put it into HttpServiceResponse(implements WriteStream);

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/BulkRecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/BulkRecordDaoImpl.java
@@ -39,6 +39,7 @@ public class BulkRecordDaoImpl implements BulkRecordDao {
           RowStream<Row> rowStream = preparedStatement.createStream(1, bindParams);
 
           rowStream.exceptionHandler(err -> {
+            transaction.close();
             System.out.println("Error: " + err.getMessage());
           });
           rowStream.endHandler(v -> {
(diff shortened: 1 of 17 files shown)


mod-source-record-storage-server/src/main/java/org/folio/dao/BulkRecordDaoImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    Promise<RowStream<Row>> promise = Promise.promise();
+    pgPool.getConnection(connectionAsyncResult -> {
+      SqlConnection connection = connectionAsyncResult.result();
+      connection.prepare(sqlQuery, preparedStatementAr -> {
+        if (preparedStatementAr.succeeded()) {
+          PreparedStatement preparedStatement = preparedStatementAr.result();
+          // Streams require to run within a transaction
+          Transaction transaction = connection.begin();
+
+          RowStream<Row> rowStream = preparedStatement.createStream(1, bindParams);
```
Igor-Gorchakov, line 39: Using "fetchSize" = 1 just to check that the stream retrieves records one-by-one

oleksandr-dekin (approved):

andrii-novytskyi (approved):

oleksandr-dekin (approved):

RuslanLavrov (approved):

TarasSpashchenko (approved):

OleksiiKuzminov (approved):

PR #288, opened 2021-02-01
MODSOURCE-173 MODSOURCE-177 Refactor data-import processing to use Kafka

KaterynaSenchenko:
## Purpose
Modules involved in data-import flow should communicate via kafka to increase reliability

## Approach
Records are processed and sent to Kafka instead of sending HTTP calls to pubsub


OleksiiKuzminov (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -37,7 +37,19 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import javax.ws.rs.BadRequestException;
```
wwelling, line 40: This added many duplicate imports in this class.

PR #292, opened 2021-02-04
MODSOURCE-219 User should be able to use a MARC-MARC match, and then update the associated Instance

OleksiiKuzminov:
## Purpose
User should be able to use a MARC-MARC match, and then update the associated Instance

## Approach
Use post-processing mechanism in data-import-processing-core to load needed entities in mod-inventory after the main marc-2-marc match operation


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

VRohach (approved):
@OleksiiKuzminov  Could you update NEWS.md, please?

PR #293, opened 2021-02-07
MODSOURCE-239: Upgrade to RAML Module Builder 32.x

VRohach:
## Purpose
The main purpose of this PR is to upgrade the module to the new version of RMB 32.x and Vertx 4.
Moreover, there are some library updates for this change. 
In that libraries, where API was changed, implementation was improved.

## Approach
All changes which are needed to support RMB 32.x and Vertx 4.x were applied.

## Learning
More info: https://issues.folio.org/browse/MODSOURCE-239


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -707,7 +716,7 @@ private ReactiveClassicGenericQueryExecutor getQueryExecutor(String tenantId) {
     return postgresClientFactory.getQueryExecutor(tenantId);
   }
 
-  private PgPool getCachecPool(String tenantId) {
+  private PgPool getCachePool(String tenantId) {
```
wwelling, line 719: Thanks for finding the typo. It should however be `getCachedPool`.

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -172,13 +176,15 @@ public RecordDaoImpl(final PostgresClientFactory postgresClientFactory) {
       .offset(offset)
       .limit(limit)
       .getSQL(ParamType.INLINED);
-    return getCachecPool(tenantId).rxBegin()
+
+    return getCachePool(tenantId)
```
wwelling, line 180: This is an incomplete conversion. This is no longer running in a transaction. Following the examples here https://vertx.io/docs/vertx-pg-client/java/#_streaming, it should be:

```
    return getCachedPool(tenantId)
      .rxGetConnection()
      .flatMapPublisher(conn -> conn.rxBegin()
        .flatMapPublisher(tx -> conn.rxPrepare(sql)
          .flatMapPublisher(pq -> pq.createStream(1)
            .toFlowable()
            .map(this::toRow)
            .map(this::toRecord))
          .doAfterTerminate(tx::commit)));
```

This affords to fetch the cursor with a single row and stream the remaining.

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -443,13 +449,16 @@ public RecordDaoImpl(final PostgresClientFactory postgresClientFactory) {
...
```
wwelling, line 453: Same as above, no longer in a transaction. Should be:

```
    return getCachedPool(tenantId)
      .rxGetConnection()
      .flatMapPublisher(conn -> conn.rxBegin()
        .flatMapPublisher(tx -> conn.rxPrepare(sql)
          .flatMapPublisher(pq -> pq.createStream(1)
            .toFlowable()
            .map(this::toRow)
            .map(this::toSourceRecord))
          .doAfterTerminate(tx::commit)));
```

wwelling (changes requested):
Good upgrade and cleanup. Two changes requested; typo correction should be to `getCachedPool` and to stream results in a transaction.

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
VRohach: @wwelling done

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
VRohach: @wwelling done

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
VRohach: @wwelling done

VRohach:
> Good upgrade and cleanup. Two changes requested; typo correction should be to `getCachedPool` and stream results should be in a transaction affording fetching single result and continue streaming with cursor.

@wwelling  Thank you for your comments! I changed that, could you please double-check?

wwelling (approved):
Approved pending RMB upgrade of mod-pubsub to fix test and passing SonarQube.

descriptors/ModuleDescriptor-template.json
```diff
@@ -246,20 +246,30 @@
...
             "pubsub.subscribers.post"
           ]
+        },
+        {
+          "methods": [
+            "DELETE"
+          ],
+          "pathPattern": "/_/tenant/{id}",
+          "modulePermissions": [
+            "pubsub.event-types.post",
```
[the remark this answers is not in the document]
VRohach: @KaterynaSenchenko  Thanks. Changed.

KaterynaSenchenko (approved):

RuslanLavrov (approved):

OleksiiKuzminov (approved):

PR #308, opened 2021-02-27
Fix Unit test: Not equals : MARC != EDIFACT MODSOURCE-251

adamdickmeiss:
See https://issues.folio.org/browse/MODSOURCE-251 . Only a PR which changes tests a bit.

Problem with existing test was that it assumes all records in there were MARC. Well. They were,.. but one.


wwelling (approved):
Thanks.

PR #313, opened 2021-03-11
MODSOURCE-223 Design and implement service layer for MARC search functionality

Igor-Gorchakov:
## Purpose
https://issues.folio.org/browse/MODSOURCE-223.
The purpose of this story to turn on a search of marc record uuis by the incoming leaderSearchExpression and fieldsSearchExpression.

The incoming search expression can be written in various manners following the requirements in this user story:
https://issues.folio.org/browse/MODSOURMAN-396.
These are requirements implemented for today:
1) Allows the user to do a left-anchored search of the data in specified subfields
2) Allows the user to query the indicator values in specified fields 
3) Allows the user to query control fields, including the leader
4) Allows the user to combine queries with boolean operators

The search is exposed via requests to an API endpoint
By default the search returns only records that are not suppressed from discovery

## Approach
_How does this change fulfill the purpose?_

#### TODOS and Open Questions

## Learning

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
@@ -130,13 +130,13 @@ public class RecordServiceImpl implements RecordService {
   }
 
   @Override
-  public Flowable<String> streamMarcRecordIds(String leaderExpression, String fieldsExpression, Boolean deleted, int offset, int limit, String tenantId) {
+  public Flowable<String> streamMarcRecordIds(String leaderExpression, String fieldsExpression, Boolean deleted, Boolean suppress, int offset, int limit, String tenantId) {
     if (leaderExpression == null && fieldsExpression == null) {
       throw new IllegalArgumentException("The 'leaderSearchExpression' and the 'fieldsSearchExpression' are missing");
     }
     ParseLeaderResult parseLeaderResult = SearchExpressionParser.parseLeaderSearchExpression(leaderExpression);
     ParseFieldsResult parseFieldsResult = SearchExpressionParser.parseFieldsSearchExpression(fieldsExpression);
-    return recordDao.streamMarcRecordIds(parseLeaderResult, parseFieldsResult, deleted, offset, limit, tenantId);
+    return recordDao.streamMarcRecordIds(parseLeaderResult, parseFieldsResult, deleted, suppress, offset, limit, tenantId);
   }
 
   @Override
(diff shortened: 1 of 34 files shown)


mod-source-record-storage-server/src/test/java/org/folio/services/SearchExpressionParserUnitTest.java
```diff
@@ -0,0 +1,365 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(BlockJUnit4ClassRunner.class)
+public class SearchExpressionParserUnitTest {
+
+  /* - TESTING SearchExpressionParser#parseFieldsSearchExpression */
```
andrii-novytskyi, line 26: do we need this comment?

illia-borysenko-ebsco (approved):

andrii-novytskyi (approved):

andrii-novytskyi:
looking good for me, but hard to understand what is going on with sonar

mod-source-record-storage-server/src/test/java/org/folio/services/SearchExpressionParserUnitTest.java
Igor-Gorchakov: yes

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -126,6 +129,16 @@ public RecordServiceImpl(final RecordDao recordDao) {
     return recordDao.streamSourceRecords(condition, recordType, orderFields, offset, limit, tenantId);
   }
 
+  @Override
+  public Flowable<String> streamMarcRecordIds(String leaderExpression, String fieldsExpression, Boolean deleted, Boolean suppress, int offset, int limit, String tenantId) {
+    if (leaderExpression == null && fieldsExpression == null) {
```
wwelling, line 134: Shouldn't this be an or? Currently, either one can be non-null and it would continue. I see the following methods perform a null check as well. Also, should it check for empty string as well? `StringUtils.isEmpty`

wwelling (approved):

illia-borysenko (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
Igor-Gorchakov: We throw an exception only when both leaderExpression & fieldsExpression are null. If one of them is not null we just continue the request processing.
There is a validation in place. There are validation rules
https://github.com/folio-org/mod-source-record-storage/pull/313/files#diff-023247c5abe9543eb4d7d877eeb48e32ce1ff7702bfeaa9567e83b6954a22e9c

RuslanLavrov (approved):

PR #330, opened 2021-05-04
MODSOURCE-269 Support search by indicator presence

psmagin:
## Purpose
Implement presence or absence searches for indicators

## Approach
* Add support of `is 'present/absent'` in `IndicatorBinaryOperand`. The empty indicator represents the `#` symbol in the database.

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/util/parser/lexeme/operand/PresenceBinaryOperand.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/util/parser/lexeme/operand/PresenceBinaryOperand.java
@@ -10,6 +10,7 @@ import static java.lang.String.format;
  * 010.a is 'absent'            - check sub field for absence
  */
 public class PresenceBinaryOperand {
+
   private static final String PRESENT = "present";
   private static final String ABSENT = "absent";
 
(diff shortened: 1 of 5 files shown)


Igor-Gorchakov:
Please, update news.md, put your Jira there

Igor-Gorchakov:
Are you sure this is a correct syntax `050.ind1 is 'present'` ? Librarians may want to check the presence of indicators in a concrete subfield, not in a marc field

psmagin:
> Are you sure this is a correct syntax `050.ind1 is 'present'` ? Librarians may want to check the presence of indicators in a concrete subfield, not in a marc field

As I know indicators are related to the full field and not to subfields as you can see on the screenshot below
image

psmagin:
> Please, update NEWS.md, put your Jira there

NEWS updated

Igor-Gorchakov:
> > Are you sure this is a correct syntax `050.ind1 is 'present'` ? Librarians may want to check the presence of indicators in a concrete subfield, not in a marc field
> 
> @Igor-Gorchakov, as I know indicators are related to the full field and not to subfields as you can see on the screenshot below
> image

Yes, you are right, indicators are stored per the field, just double-checked

mod-source-record-storage-server/src/main/java/org/folio/services/util/parser/lexeme/operand/PresenceBinaryOperand.java
```diff
@@ -10,6 +10,7 @@
  * 010.a is 'absent'            - check sub field for absence
```
Igor-Gorchakov, line 10: Update javadoc, add your cases, please

psmagin: Updated

Igor-Gorchakov (approved):

PR #335, opened 2021-05-12
MODSOURCE-288 Migrate QM-flow to Kafka

psmagin:
## Purpose
Migrate QM record update flow to Kafka.

## Approach
* Support consuming `QM_RECORD_UPDATED` event
* Support producing `QM_SRS_MARC_RECORD_UPDATED` and `QM_ERROR` events
* Deploying new verticle for consumer


Igor-Gorchakov (approved):

mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/QuickMarcConsumersVerticle.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static AbstractApplicationContext springGlobalContext;
+
+  @Autowired
+  private QuickMarcKafkaHandler kafkaHandler;
+
+  @Autowired
+  private KafkaConfig kafkaConfig;
+
+  @Value("${srs.kafka.QuickMarcConsumer.loadLimit:5}")
```
KaterynaSenchenko, line 31: please add a note in Readme file that these parameters will be required for the Juniper release. See https://github.com/folio-org/mod-source-record-storage/pull/338

KaterynaSenchenko (approved):

natali-zaitseva (approved):

mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/QuickMarcConsumersVerticle.java
psmagin: added

PR #340, opened 2021-05-25
MODSOURCE-301 - Cannot import GOBI EDIFACT invoice

RuslanLavrov:
## Purpose
to import an EDIFACT invoice for GOBI. StAEDI library recognizes `31B` code for interchange sender qualifier (in UNB segment) as invalid but according to EDItEUR EDI implementation guideline it's expected value _(31B = US SAN Agency (EDItEUR code))_.

## Approach
* add `31B` code to exceptions as expected value (set EDI_VALIDATE_CONTROL_CODE_VALUES property to false, to allow code values that are not included to validation schema of staedi library)
* verify that parsed content is not null on normalization before event sending
* update tests

## Learning
[MODSOURCE-301](https://issues.folio.org/browse/MODSOURCE-301)
[EDItEUR EDIFACT guideline](https://www.editeur.org/files/EDIfact%20eancom%20pdfs/codes%20draft%20feb%202010.pdf) (page 5, Partner identification code qualifier)


KaterynaSenchenko (approved):
update the news pls

PR #342, opened 2021-06-01
MODSOURCE-310: Handling 001/003/035 handling in SRS for MARC bib records broken

VRohach:
## Purpose
Fix logic for 001/003/035 handling for overlaying.

## Approach

- For instance=MARC: was added mechanism for removing 003 field, if it`s needed.
- For instance=Folio: 001/003/035 logic wasn't applied because ExternalHoldersIds wasn't empty(before we have check, and don`t execute it if ExternalHoldersIds is not empty). **This bug caused that the updated instance can't be saved in mod-inventory, because the new instance has an invalid hrid** (ocm12345, for example, not in0000001). So, in this PR was added a condition, that if instance.id==record.externalHolderIds.instanceId AND  instace.hrid!=record.externalHolderIds.instanceHrid(in this case it reveales, ocm1234!=in0000001), then this mechanism will be executed. This specific case (for not equals between marc-record and instance) was in this bug. 

## Learning
More info: https://issues.folio.org/browse/MODSOURCE-310

**NOTE: Tests are in progress!**

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 * [MODSOURCE-276](https://issues.folio.org/browse/MODSOURCE-276) Add existing records to the SRS Query API table
 * [MODSOURCE-288](https://issues.folio.org/browse/MODSOURCE-288) Migrate QM-flow to Kafka
 * [MODSOURCE-279](https://issues.folio.org/browse/MODSOURCE-279) Store MARC Authority record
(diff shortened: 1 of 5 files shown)


VRohach:
Guys, tests are in progress. I`m working on them now.

VRohach:
Feel free to review it. Thanks!

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

PR #345, opened 2021-06-04
MODSOURCE-311 Search API: Restrict to search only by Marc Bib

Igor-Gorchakov:
## Purpose
https://issues.folio.org/browse/MODSOURCE-311
The HTTP response contains empty strings when the search includes Marc Authorities

## Approach
The Search API should be restricted to do search only by the Marc Bibs

- Queries are restricted in a 'where' clause
- Created a class 'RecordSearchParameters'. This is a class for parameters needed for the search. The HTTP request will contain more than 7 parameters later, so it makes sense to use such a class-holder to satisfy Sonar Cube and have more pretty method signature :)

#### TODOS and Open Questions

## Learning


mod-source-record-storage-server/src/main/java/org/folio/services/RecordSearchParameters.java
```diff
@@ -0,0 +1,88 @@
...
+
+import org.folio.rest.jaxrs.model.MarcRecordSearchRequest;
+import org.folio.rest.jaxrs.model.Record;
+
+/**
+ * Parameters needed for extended search
+ *
+ * @see MarcRecordSearchRequest
+ */
+public class RecordSearchParameters {
```
psmagin, line 11: Using Lombok will help to decrease the amount of boilerplate code. Also will be great to have equals and hashcode overridden.

mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SourceStorageStreamApiTest.java
```diff
@@ -1204,6 +1204,54 @@ public void shouldReturnEmptyResponseOnSearchMarcRecordIdsWhenOffsetIs1(TestCont
     async.complete();
   }
 
+  @Test
+  public void shouldReturnIdResponseOnSearchMarcRecordIdsWhenMarcBibAndAuthoritySaved(TestContext testContext) {
+    // given
+    final Async async = testContext.async();
+    postSnapshots(testContext, snapshot_2);
+    postRecords(testContext, record_2, record_7);
```
psmagin, line 1212: Will be great to rename record_2 and record_7 to another that represent type of the record. For example: auth_record_2, bib_record_1.
It will improve the readability of tests.

mod-source-record-storage-server/src/main/java/org/folio/services/RecordSearchParameters.java
```diff
@@ -0,0 +1,88 @@
...
+  private String leaderSearchExpression;
+  private String fieldsSearchExpression;
+  private Record.RecordType recordType;
+  private boolean deleted;
+  private boolean suppressedFromDiscovery;
+  private Integer limit;
+  private Integer offset;
+
+  public static RecordSearchParameters from(MarcRecordSearchRequest request) {
```
psmagin, line 20: Will be good to have notnull check for `request`

mod-source-record-storage-server/src/main/java/org/folio/services/RecordSearchParameters.java
Igor-Gorchakov: @KaterynaSenchenko  does it make sense to use the Lombok library for this class to make it pretty?

KaterynaSenchenko: @Igor-Gorchakov, I have no objections, you can use Lombok

oleksandr-dekin (approved):

RuslanLavrov (approved):

PR #353, opened 2021-06-22
MODSOURCE-323 - Change dataType to have have common type for MARC related subtypes

natali-zaitseva:
## Purpose
Fix for  https://issues.folio.org/browse/MODSOURCE-323

## Approach
* changed the reference to submodule
* changed the data type in test


natali-zaitseva:
should be updated to latest master for submodules after folio-org/data-import-raml-storage#217 merged

KaterynaSenchenko (approved):

PR #368, opened 2021-08-17
MODSOURCE-351 Endpoint to verify invalid MARC Bib ids in the system

oleksandr-dekin:
## Purpose
Search MARC BIB ids in the system and return invalid HRID 

## Approach
- Create endpoint for searching invalid MARC BIb ids
- Extend or create new raml for endpoint
- Create DTO for returning response into SRM
- Implement DAO and service layer
- Сhange module descriptor
- Update new version
- Write tests

## Learning
https://issues.folio.org/browse/MODSOURCE-351

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -195,7 +195,7 @@
     },
     {
       "id": "source-storage-batch",
-      "version": "1.0",
+      "version": "1.1",
       "handlers": [
         {
           "methods": [
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -214,6 +214,15 @@
           "permissionsRequired": [
             "source-storage.records.put"
           ]
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/source-storage/batch/verified-records",
+          "permissionsRequired": [
+            "source-storage.verified.records"
+          ]
         }
       ]
     },
@@ -294,6 +303,11 @@
       "description": "Get Results"
     },
     {
+      "permissionName": "source-storage.verified.records",
+      "displayName": "Source Storage - validate marc bib ids in the system",
+      "description": "Return marc bib ids, which doesn't exist in the system"
+    },
+    {
       "permissionName": "source-storage.all",
       "displayName": "Source Record Storage - all permissions",
       "description": "Entire set of permissions needed to manage snapshots and records",
@@ -308,7 +322,8 @@
         "source-storage.records.put",
         "source-storage.records.delete",
         "source-storage.record.update",
-        "source-storage.sourceRecords.get"
+        "source-storage.sourceRecords.get",
+        "source-storage.verified.records"
       ],
       "visible": false
     }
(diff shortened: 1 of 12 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -214,6 +214,15 @@
...
           ]
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/source-storage/batch/verified-records",
+          "permissionsRequired": [
+            "source-storage.verified.records"
+          ]
```
psmagin, line 225: Also, interface version should be updated.

mod-source-record-storage-server/src/main/java/org/folio/services/RecordService.java
```diff
@@ -189,4 +191,13 @@
    */
   Future<Record> updateSourceRecord(ParsedRecordDto parsedRecordDto, String snapshotId, String tenantId);
 
+  /**
+   * Find marc bib ids by incoming arrays from SRM and exclude all valid marc bib and return only marc bib ids,
+   * which does not exists in the system
+   *
+   * @param marcBibIds list of invalid marc bib ids
```
psmagin, line 198:
```suggestion
   * @param marcBibIds list of marc bib ids
```

mod-source-record-storage-server/src/main/java/org/folio/services/RecordService.java
```diff
@@ -189,4 +191,13 @@
...
+   * @param tenantId tenant id
+   * @return future with list of invalid marc bib ids
+   */
+  Future<MarcBibCollection> verifyMarcBibRecords(List<String> marcBibIds, String tenantId);
```
psmagin, line 202: I think that `MarcBibCollection`naming is confusing because the only that we return is IDs of MARC_BIB records that were not found in the database.

mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SourceStorageBatchApiTest.java
```diff
@@ -59,6 +65,7 @@
   private static final String THIRD_UUID = UUID.randomUUID().toString();
   private static final String FOURTH_UUID = UUID.randomUUID().toString();
   private static final String FIFTH_UUID = UUID.randomUUID().toString();
+  public static final String VALID_HRID = "12345";
```
psmagin, line 68:
```suggestion
  private static final String VALID_HRID = "12345";
```

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -763,6 +767,34 @@ private void appendWhere(SelectJoinStep step, ParseLeaderResult parseLeaderResul
...
+      return Future.succeededFuture(new MarcBibCollection());
+    }
+    var marcHrid = DSL.field("marc.hrid");
+
+    return getQueryExecutor(tenantId).transaction(txQE -> txQE.query(dsl ->
+      dsl.selectDistinct(marcHrid)
+        .from("(SELECT unnest(" + DSL.array(marcBibIds.toArray()) + ") as hrid) as marc")
+        .leftJoin(RECORDS_LB)
+        .on(RECORDS_LB.INSTANCE_HRID.eq(marcHrid.cast(String.class)))
+        .where(RECORDS_LB.INSTANCE_HRID.isNull())
```
psmagin, line 782: Could you please modify the query by including condition of `recordType == MARC_BIB`

KaterynaSenchenko (approved):

natali-zaitseva (approved):

psmagin (approved):

PR #370, opened 2021-08-27
MODSOURCE-342 Add HoldingsPostProcessingEventHandler

psmagin:
## Purpose
Create handler for post processing event

## Approach
* Create AbstractPostProceesing handler
* Create 2 implementations: for instance and for holdings


mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -372,14 +370,25 @@ public static void remove003FieldIfNeeded(Record record, String instanceHrid) {
...
+   *
+   * @param record         - source record.
+   * @param externalEntity - source externalEntity.
    * @return - true if need.
    */
-  public static boolean ifFillingFieldsNeeded(Record record, JsonObject instance) {
-    return (StringUtils.isNotEmpty(record.getExternalIdsHolder().getInstanceId()) && StringUtils.isNotEmpty(record.getExternalIdsHolder().getInstanceHrid())) &&
-      (instance.getString(ID_FIELD).equals(record.getExternalIdsHolder().getInstanceId())
-        && !instance.getString(HR_ID_FIELD).equals(record.getExternalIdsHolder().getInstanceHrid()));
+  public static boolean ifFillingFieldsNeeded(Record record, JsonObject externalEntity) {
```
natali-zaitseva, line 378: could you rename method to isFieldsFillingNeeded to align with already existing method names such as isNotEmpty(), isEligible(), isPresent()

natali-zaitseva (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
psmagin: Renamed

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -372,14 +370,25 @@ public static void remove003FieldIfNeeded(Record record, String instanceHrid) {
...
-  public static boolean ifFillingFieldsNeeded(Record record, JsonObject instance) {
-    return (StringUtils.isNotEmpty(record.getExternalIdsHolder().getInstanceId()) && StringUtils.isNotEmpty(record.getExternalIdsHolder().getInstanceHrid())) &&
-      (instance.getString(ID_FIELD).equals(record.getExternalIdsHolder().getInstanceId())
-        && !instance.getString(HR_ID_FIELD).equals(record.getExternalIdsHolder().getInstanceHrid()));
+  public static boolean isFieldsFillingNeeded(Record record, JsonObject externalEntity) {
+    var recordType = record.getRecordType();
+    var externalIdsHolder = record.getExternalIdsHolder();
+    var id = externalEntity.getString(ID_FIELD);
+    var hrid = externalEntity.getString(HR_ID_FIELD);
+    if (Record.RecordType.MARC_BIB == recordType) {
```
oleksandr-dekin, line 383: Maybe better to introduce validation as method? Like this:
```suggestion
    if (Record.RecordType.MARC_BIB == recordType) {
      return isValidIdAndHrid(id, hrid, externalIdsHolder.getInstanceId(), externalIdsHolder.getInstanceHrid());
    }
    if (Record.RecordType.MARC_HOLDING == recordType) {
      return isValidIdAndHrid(id, hrid, externalIdsHolder.getHoldingsId(), externalIdsHolder.getHoldingsHrid());
    }
    return false;
  }

  private static boolean isValidIdAndHrid(String id, String hrid, String instanceId, String instanceHrid) {
    return (isNotEmpty(instanceId) && isNotEmpty(instanceHrid))
      && (id.equals(instanceId) && !hrid.equals(instanceHrid));
  }
```

oleksandr-dekin (approved):

PR #376, opened 2021-09-08
MODSOURCE-286 - Remove zipping mechanism for data import event payloads and use cache for params

RuslanLavrov:
## Purpose
to reduce event payload size by removing jobProfileSnapshot and mappingParameters from the payload

## Approach
* load jobProfileSnapshot from mod-data-import-converter-storage and add cache to store it for subsequent usage
* load mappingParameters from mod-source-record-manager and add cache to store them for subsequent usage
* remove all zipping/unzipping operations for data-import payload
* add and update tests

## Learning
[MODSOURCE-286](https://issues.folio.org/browse/MODSOURCE-286)
should be merged after https://github.com/folio-org/data-import-processing-core/pull/166


bisecomsergiy (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
```diff
@@ -71,11 +71,12 @@ private EventHandlingUtil() { }
   public static KafkaProducerRecord<String, String> createProducerRecord(String eventPayload, String eventType,
```
VRohach, line 71: @RuslanLavrov  I think it would be better to change this method. Remove the last argument and remove zipping logic inside this one.

VRohach (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/ModifyRecordEventHandler.java
```diff
@@ -145,4 +164,12 @@ public boolean isPostProcessingNeeded() {
   public String getPostProcessingInitializationEventType() {
     return DI_SRS_MARC_BIB_RECORD_MODIFIED_READY_FOR_POST_PROCESSING.value();
   }
+
+  private OkapiConnectionParams retrieveOkapiConnectionParams(DataImportEventPayload eventPayload) {
```
SerhiiNosko, line 168: We can use RestUtil.retrieveOkapiConnectionParams here

mod-source-record-storage-server/src/main/java/org/folio/services/util/EventHandlingUtil.java
RuslanLavrov: @VRohach, no, to avoid breaking changes since this method also is used by quick mark editor functionality that uses compressed payload https://github.com/folio-org/mod-source-record-storage/blob/037c6211ba27e4448513ca841d290ff87dffa5cd/mod-source-record-storage-server/src/main/java/org/folio/services/QuickMarcKafkaHandler.java#L121

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/ModifyRecordEventHandler.java
RuslanLavrov: sure, just turned out this refactoring has not been pushed yet, thanks!

SerhiiNosko (approved):

KaterynaSenchenko (approved):

PR #396, opened 2021-11-18
MODSOURCE-402 Properly handle DB failures during events processing

Abdulkhakimov:
## Purpose
Prevent sending events from srs during duplicate event

## Approach
- Finding a duplicate event by throwing a unique constraint violation
- Throws a custom exception and handles in ParsedRecordChunksErrorHandler (only output to logs, do not send DI_ERROR)

## Learning
https://issues.folio.org/browse/MODSOURCE-402

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
@@ -12,6 +12,7 @@ import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.folio.dao.exceptions.DuplicateEventException;
 import org.folio.dao.util.ErrorRecordDaoUtil;
 import org.folio.dao.util.IdType;
 import org.folio.dao.util.ParsedRecordDaoUtil;
@@ -118,6 +119,7 @@ public class RecordDaoImpl implements RecordDao {
 
   private static final int DEFAULT_LIMIT_FOR_GET_RECORDS = 1;
 
+  private static final String UNIQUE_VIOLATION_SQL_STATE = "23505";
   private static final String RECORD_NOT_FOUND_BY_ID_TYPE = "Record with %s id: %s was not found";
   private static final String INVALID_PARSED_RECORD_MESSAGE_TEMPLATE = "Record %s has invalid parsed record; %s";
 
@@ -439,6 +441,9 @@ public class RecordDaoImpl implements RecordDao {
           .errors();
 
         recordsLoadingErrors.forEach(error -> {
+          if(error.exception().sqlState().equals(UNIQUE_VIOLATION_SQL_STATE)) {
+            throw new DuplicateEventException("Error when handling duplicate event");
+          }
           LOG.warn("Error occurred on batch execution: {}", error.exception().getCause().getMessage());
           LOG.debug("Failed to execute statement from batch: {}", error.query());
         });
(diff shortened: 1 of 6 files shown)


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -439,6 +441,9 @@ private void appendWhere(SelectJoinStep step, ParseLeaderResult parseLeaderResul
           .errors();
 
         recordsLoadingErrors.forEach(error -> {
+          if(error.exception().sqlState().equals(UNIQUE_VIOLATION_SQL_STATE)) {
+            throw new DuplicateEventException("Error when handling duplicate event");
```
SerhiiNosko, line 445: Can we add some identifiers of duplicate events in message such as jobExecutionId, recordId etc?

mod-source-record-storage-server/src/main/java/org/folio/errorhandlers/ParsedRecordChunksErrorHandler.java
```diff
@@ -57,7 +58,11 @@ public void handle(Throwable throwable, KafkaConsumerRecord<String, String> reco
     String jobExecutionId = okapiConnectionParams.getHeaders().get(JOB_EXECUTION_ID_HEADER);
     String tenantId = okapiConnectionParams.getTenantId();
 
-    sendErrorRecordsSavingEvents(recordCollection, throwable.getMessage(), kafkaHeaders, jobExecutionId, tenantId);
+    if(throwable instanceof DuplicateEventException) {
+      LOGGER.warn("Error parsing event for jobExecutionId: {} , tenantId: {}, cause: {}", jobExecutionId, tenantId, throwable.getMessage());
```
SerhiiNosko, line 62: The text of this log message should be changed to someone like this: "Duplicate event received, skipping processing" and include jobExecutionId, chunkId or recordId

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
Abdulkhakimov: Unfortunately, it will not succeed. Only record Collection and TenantId are in the method arguments.

mod-source-record-storage-server/src/main/java/org/folio/errorhandlers/ParsedRecordChunksErrorHandler.java
Abdulkhakimov: Thanks. I forgot to fix the temporary draft text. We can't add recordId because there can be multiple records. We don't have access to chunkNumber from ErrorHandler.

RuslanLavrov: hello Khamid, please add "correlationId" header to log message (okapiConnectionParams.getHeaders().get("correlationId")) to track what chunk/event was processed repeatedly

RuslanLavrov (approved):

SerhiiNosko (approved):

mod-source-record-storage-server/src/main/java/org/folio/errorhandlers/ParsedRecordChunksErrorHandler.java
Abdulkhakimov: > hello Khamid, please add "correlationId" header to log message (okapiConnectionParams.getHeaders().get("correlationId")) to track what chunk/event was processed repeatedly

Hello Ruslan. Thank you for comment. Done

VRohach (approved):

PR #411, opened 2022-01-05
MODSOURCE-433 Implement update Authority handler

psmagin:
## Purpose
Implement update Authority handler

## Approach
* Extract AbstractUpdateModifyEventHandler that is a base for handling records update.
* Implement handlers for bib and authority records.


Shans-Kaluhin (approved):
Looks excellent!

PR #419, opened 2022-01-26
MODSOURCE-446 Extend MARC-MARC matching to 9xx and 0xx fields

Igor-Gorchakov:
## Purpose

https://issues.folio.org/browse/MODSOURCE-446
The purpose is to allow matching the incoming imported Marc record to the existing SRS Marc record by any Marc field(001 - 999) that is specified in the matching criteria of a Match profile

## Approach

The approach is to implement a DAO method that searches Records by the given Marc field.
The method builds an SQL query to select Records from _records_lb_ table joined to the concrete partition of _marc_indexers_ table. The partition is associated with a given Marc field being matched.

If there is a need to match by Control field <marc_field> = '12345' the SQL query will look like:
```
select from records_lb
inner join marc_indexers_<marc_field> on records_lb.id = marc_indexers_<marc_field>.marc_id
where marc_indexers_<marc_field>.value = '12345'
```
If there is a need to match Data field <marc_field><indicator1><indicator2>$<subfield> = '30a7fc76-7067-4c98-a904-c5d14f2ac0b8' the SQL query will look like:
```
select from records_lb
inner join marc_indexers_<marc_field> on records_lb.id = marc_indexers_<marc_field>.marc_id
where marc_indexers_<marc_field>.value = '30a7fc76-7067-4c98-a904-c5d14f2ac0b8' and marc_indexers_<marc_field>.ind1 = '<indicator1>' and  marc_indexers_<marc_field>.ind2 = '<indicator2>' and  marc_indexers_<marc_field>.subfield_no = '<subfield>'
```

## Learning
Screenshot 2022-01-25 at 14 50 27


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -176,6 +185,51 @@ public RecordDaoImpl(final PostgresClientFactory postgresClientFactory) {
     )).map(queryResult -> toRecordCollectionWithLimitCheck(queryResult, limit));
   }
 
+  public Future<RecordCollection> getMatchedRecords(MatchField matchedField, TypeConnection typeConnection, int offset, int limit, String tenantId) {
```
Igor-Gorchakov, line 188: This method is written similar to the existing method _getRecords()_ that was used earlier. I just added an innerJoin to the partition of _marc_indexers_ table

psmagin (approved):

KaterynaSenchenko (approved):

RuslanLavrov (approved):

PR #432, opened 2022-02-16
MODSOURCE-457 Delete Authority

Igor-Gorchakov:
## Purpose
_Describe the purpose of the pull request. Include background information if necessary._

## Approach
_How does this change fulfill the purpose?_

#### TODOS and Open Questions

## Learning


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -879,6 +879,13 @@ private MarcBibCollection toMarcBibCollection(QueryResult result) {
     return SnapshotDaoUtil.delete(getQueryExecutor(tenantId), snapshotId);
   }
 
+  @Override
+  public Future<Boolean> deleteRecordById(String recordId, String tenantId) {
+    return getQueryExecutor(tenantId).execute(dsl -> dsl.deleteFrom(RECORDS_LB)
+        .where(RECORDS_LB.ID.eq(UUID.fromString(recordId))))
+      .map(res -> res == 1);
+  }
```
psmagin, line 887: It was decided to use a soft-delete approach for deleting record in SRS. It means that the record should have state = DELETED in the database.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
```diff
@@ -0,0 +1,102 @@
...
+      return future;
+    } else {
+      payload.getEventsChain().add(payload.getEventType());
+      var record = Json.decodeValue(payloadContext.get(getRecordKey()), Record.class);
+      deleteRecord(record, payload.getTenant())
+        .onSuccess(isDeleted -> {
+          if (isDeleted) {
+            payload.setEventType(getNextEventType());
+            payload.getContext().remove(getRecordKey());
+            payload.getContext().put(getRecordIdKey(), record.getId());
```
psmagin, line 60: Payload should have not SRS record ID but a related record ID, in the current ticket scope, it is Inventory Authority ID. You could retrieve it from the ExternalIdsHolder field in the record.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/MarcAuthorityDeleteEventHandler.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Autowired
+  public MarcAuthorityDeleteEventHandler(RecordService recordService) {
+    super(recordService, TypeConnection.MARC_AUTHORITY);
+  }
+
+  @Override
+  protected Future<Boolean> deleteRecord(Record record, String tenantId) {
+    return recordService.deleteRecordById(record.getId(), tenantId);
+  }
```
psmagin, line 23: Could you please explain your decision to make this method abstract and overrode? I think that approach of deleting record will be the same for all type of records.

psmagin (changes requested):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
Igor-Gorchakov: It's just an old changes, will push the latest soon

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /* Handles DELETE action  */
+  private void handlePayload(DataImportEventPayload payload, CompletableFuture<DataImportEventPayload> future) {
+    var record = Json.decodeValue(payload.getContext().get(getRecordKey()), Record.class);
+    recordService.updateRecord(record.withState(Record.State.DELETED), payload.getTenant())
+      .onSuccess(ar -> {
+        payload.setEventType(getNextEventType());
+        payload.getContext().remove(getRecordKey());
+        payload.getContext().put(getExternalRecordIdKey(), record.getExternalIdsHolder().getAuthorityId());
```
psmagin, line 69: `record.getExternalIdsHolder().getAuthorityId()` should be extracted to abstract method and overridden in `MarcAuthorityDeleteEventHandler`. This is specific to authority flow.

ALiemeshka (approved):

psmagin (approved):

PR #436, opened 2022-02-23
MODSOURCE-459 Determine how to handle deleted authority records

Igor-Gorchakov:
## Purpose
The story supports a hard delete when a user deletes a record from the MARC authority app OR figures out a way to purge this type of delete record workflow

## Approach
The requirements say that a hard delete should be supported. But the deletion during the data-import process could be time-critical for the system.
The solution is to init a Vertx Job, that gets run by the fixed schedule (24hours delay by default). The Job runs and purges records on daily basis. The only 'DELETED' and related 'OLD' records are getting purged from SRS.

#### TODOS and Open Questions

## Learning


ALiemeshka (approved):

viacheslavkol (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -879,6 +879,22 @@ private MarcBibCollection toMarcBibCollection(QueryResult result) {
...
+  public Future<Void> cleanRecords(String tenantId) {
+    Promise<Void> promise = Promise.promise();
+    var purgeOldRecordsQuery = DSL.deleteFrom(RECORDS_LB)
+      .where(RECORDS_LB.STATE.eq(RecordState.OLD).and(RECORDS_LB.MATCHED_ID.in(DSL.select(RECORDS_LB.ID).from(RECORDS_LB).where(RECORDS_LB.STATE.eq(RecordState.DELETED)))));
+    var purgeDeletedRecordsQuery = DSL.deleteFrom(RECORDS_LB)
+      .where(RECORDS_LB.STATE.eq(RecordState.DELETED));
+    ReactiveClassicGenericQueryExecutor executor = getQueryExecutor(tenantId);
+    Future.succeededFuture()
+      .compose(ar -> executor.execute(dsl -> purgeOldRecordsQuery))
+      .compose(ar -> executor.execute(dsl -> purgeDeletedRecordsQuery))
```
RuslanLavrov, line 892: hi @Igor-Gorchakov, I propose to perform records' deletion operations in transaction

RuslanLavrov (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDao.java
```diff
@@ -287,6 +287,14 @@
    */
   Future<Boolean> deleteRecordsBySnapshotId(String snapshotId, String tenantId);
 
+  /**
+   *  Performs purge the 'DELETED' records, and its related 'OLD' records
+   *
+   * @param tenantId tenant id
+   * @return future
+   */
+  Future<Void> cleanRecords(String tenantId);
```
psmagin, line 296: It's needed to keep soft-deleted records for some time. In the approach described by SA, he suggests removing records that are older than 7 days. Also, he suggests limiting the maximum amount of records deleted for example by 500 records in a time.

psmagin (changes requested):
Please review once again approach suggest by SA.

PR #437, opened 2022-03-03
MODDATAIMP-645 fix update a MARC authority record multiple times

ALiemeshka:
## Purpose
Error while importing file with expected update and create:
io.vartex.pgclient.PgException: ERROR:  current transaction is aborted, commands ignored until end of transaction block(25P02)

## Learning
[MODDATAIMP-645](https://issues.folio.org/browse/MODDATAIMP-645)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
@@ -2,6 +2,7 @@ package org.folio.services.handlers.actions;
 
 import static java.lang.String.format;
 import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 import static org.folio.ActionProfile.Action.MODIFY;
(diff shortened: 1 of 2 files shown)


Shans-Kaluhin (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
```diff
@@ -184,6 +185,8 @@ private void preparePayload(DataImportEventPayload dataImportEventPayload) {
 
   private void increaseGeneration(Record changedRecord) {
     var generation = changedRecord.getGeneration();
-    changedRecord.setGeneration(isNull(generation) ? 1 : ++generation);
+    if(nonNull(generation)){
```
natali-zaitseva, line 188: What will happen if generation is NULL?

ALiemeshka: Nothing, just like here org.folio.services.handlers.InstancePostProcessingEventHandler#sendEventToDataImportLog

This method was added for in MODSOURMAN-692

natali-zaitseva (approved):

PR #442, opened 2022-03-17
MODSOURCE-477 Configure job to delete authority records

Igor-Gorchakov:
## Purpose
https://issues.folio.org/browse/MODSOURCE-477
Now the Job is created and runs every day and purges all the 'deleted' records by the fixed schedule. In this story, we have to apply some adjustments and configure the Job

## Approach
1) Restrict the Job to purge only records that were marked 'deleted' more than 7 days back;
2) Restrict the Job to purge only a limited pre-defined number of records;

#### TODOS and Open Questions

## Learning


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -880,12 +881,14 @@ private MarcBibCollection toMarcBibCollection(QueryResult result) {
...
-    var purgeOldRecordsQuery = DSL.deleteFrom(RECORDS_LB)
-      .where(RECORDS_LB.STATE.eq(RecordState.OLD).and(RECORDS_LB.MATCHED_ID.in(DSL.select(RECORDS_LB.ID).from(RECORDS_LB).where(RECORDS_LB.STATE.eq(RecordState.DELETED)))));
-    var purgeDeletedRecordsQuery = DSL.deleteFrom(RECORDS_LB)
-      .where(RECORDS_LB.STATE.eq(RecordState.DELETED));
+    var selectDeletedRecordIdsQuery = DSL
+      .select(RECORDS_LB.ID).from(RECORDS_LB)
+      .where(RECORDS_LB.STATE.eq(RecordState.DELETED)).and(RECORDS_LB.UPDATED_DATE.le(OffsetDateTime.now().minusDays(lastUpdatedDays)))
+      .limit(limit);
+    var purgeOldRecordsQuery = DSL.deleteFrom(RECORDS_LB).where(RECORDS_LB.STATE.eq(RecordState.OLD).and(RECORDS_LB.MATCHED_ID.in(selectDeletedRecordIdsQuery)));
+    var purgeDeletedRecordsQuery = DSL.deleteFrom(RECORDS_LB).where(RECORDS_LB.ID.in(selectDeletedRecordIdsQuery));
```
psmagin, line 891: As I can see from these queries `limit` parameter is not actually limit the total amount of records:
Records that will be removed if limit = 10:
 - 10 records with the DELETED state
 - from 0 to an unlimited amount of records with the OLD state

This may cause a problems

psmagin (changes requested):

Igor-Gorchakov:
Need to rework

PR #446, opened 2022-03-23
MODSOURCE-477 Configure job to delete authority records

Igor-Gorchakov:
## Purpose
https://issues.folio.org/browse/MODSOURCE-477
Now the Job is created and runs every day and purges all the 'deleted' records by the fixed schedule. In this story, we have to apply some adjustments and configure the Job

## Approach
- Restricted the Job to purge only records that were marked 'deleted' more than 7 days back;
- Restricted the Job to purge only a limited pre-defined number of records;
- Applied @Scheduler annotation to setup scheduler from Spring Framework

## Learning
https://www.baeldung.com/spring-scheduled-tasks


mod-source-record-storage-server/src/main/java/org/folio/services/cleanup/RecordCleanupService.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   * The method is getting run by the Spring Framework's Scheduler, at 12 am (midnight) every day by default.
+   * The execution starts automatically on the ApplicationContext setup.
+   * The schedule is defined by the cron expression, which allows defining timing in a fixed format.
+   */
+  @Scheduled(cron = "${srs.cleanup.cron.expression:0 0 0 * * ?}")
+  public void cleanup() {
+    TenantUtil.getModuleTenants(vertx)
```
psmagin, line 38: I think it will be useful to have some info message on the method start to show that cleanup job is started.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
```diff
@@ -64,7 +65,7 @@ public AbstractDeleteEventHandler(RecordService recordService, TypeConnection ty
   /* Handles DELETE action  */
   private void handlePayload(DataImportEventPayload payload, CompletableFuture<DataImportEventPayload> future) {
     var record = Json.decodeValue(payload.getContext().get(getRecordKey()), Record.class);
```
psmagin, line 67: Please add info message there also.

psmagin (changes requested):

psmagin (approved):

RuslanLavrov (approved):

PR #449, opened 2022-04-01
MODSOURCE-489 - Records that are overlaid multiple times via Inventory Single Record Import can't be overlaid after the second time

RuslanLavrov:


mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -102,7 +102,7 @@ public RecordServiceImpl(final RecordDao recordDao) {
           return recordDao.getRecordByMatchedId(txQE, record.getMatchedId())
             .compose(optionalMatchedRecord -> optionalMatchedRecord
               .map(matchedRecord -> recordDao.saveUpdatedRecord(txQE, ensureRecordForeignKeys(record.withGeneration(generation)), matchedRecord.withState(Record.State.OLD)))
-              .orElse(recordDao.saveRecord(txQE, ensureRecordForeignKeys(record.withGeneration(generation)))));
+              .orElseGet(() -> recordDao.saveRecord(txQE, ensureRecordForeignKeys(record.withGeneration(generation)))));
```
okolawole-ebsco, line 105: just curious, why this change?

SerhiiNosko (approved):

Abdulkhakimov (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
Abdulkhakimov: orElse calls the method and calculates the return value anyway, orElseGet lets us avoid this.

RuslanLavrov: the saveRecord() was being called, alone with saveUpdatedRecord(), even when a record was found by matchedId

PR #452, opened 2022-04-22
MODSOURCE-496 Single record imports show the incorrect record number in the summary log 1st column

AivarIUsupov:
## Purpose
When a record was part of a previous multi-record import, and is then updated via single record import, the wrong record number shows in the log for the Inventory single record import

## Approach
Added forced set of order when it is update action. 

## Learning
[MODSOURCE-496](https://issues.folio.org/browse/MODSOURCE-496)

## Note
There are a lot of test cases that can only be checked manually.
Also need regression testing for this ticket.

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
@@ -88,6 +88,7 @@ public abstract class AbstractUpdateModifyEventHandler implements EventHandler {
             remove003FieldIfNeeded(changedRecord, hrId);
           }
           increaseGeneration(changedRecord);
+          changedRecord.setOrder(0);
           payloadContext.put(modifiedEntityType().value(), Json.encode(changedRecord));
         })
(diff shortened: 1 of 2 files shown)


SerhiiNosko (approved):

okolawole-ebsco (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
```diff
@@ -88,6 +88,7 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload p
             remove003FieldIfNeeded(changedRecord, hrId);
           }
           increaseGeneration(changedRecord);
+          changedRecord.setOrder(0);
```
RuslanLavrov, line 91: Hello @AivarIUsupov, I assume there might be a scenario during importing a file with multiple records using profile for record update, the all updated records from the file will have order 0 and the records in the log as well. Propose to set the order (a record number) of incoming record into updated record to keep the order of records from incoming file.

RuslanLavrov (approved):

PR #457, opened 2022-05-13
MODSOURCE-499 fix eventloop blocking during batch records save

okolawole-ebsco:
## Purpose
Alleviate eventloop blocking during batch record saves

## Approach
move blocking actions to be executed by worker threads

## Learning
https://issues.folio.org/browse/MODSOURCE-499

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
@@ -708,7 +708,7 @@ public class RecordDaoImpl implements RecordDao {
   }
 
   @Override
-  public Future<ParsedRecordsBatchResponse> updateParsedRecords(RecordCollection recordCollection, String tenantId) {
+  public Future<ParsedRecordsBatchResponse> updateParsedRecords(RecordCollection recordCollection, String tenantId) { //NOSONAR
     Promise<ParsedRecordsBatchResponse> finalPromise = Promise.promise();
     Context context = Vertx.currentContext();
     if(context == null) return Future.failedFuture("updateParsedRecords must be called by a vertx thread");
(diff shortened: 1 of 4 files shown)


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -665,10 +683,28 @@ public Future<Integer> calculateGeneration(ReactiveClassicGenericQueryExecutor t
 
   @Override
   public Future<ParsedRecord> updateParsedRecord(Record record, String tenantId) {
```
SerhiiNosko, line 685: Do we need to follow the same approach for the updateParsedRecords method?

SerhiiNosko, line 685: Do we need to follow the same approach for the updateParsedRecords method?

SerhiiNosko (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
okolawole-ebsco: I agree that we should do the same. This has been added. Currently fighting with Sonarqube to get the PR in a green state.

Aliaksandr-Fedasiuk (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -665,152 +683,184 @@ public Future<Integer> calculateGeneration(ReactiveClassicGenericQueryExecutor t
...
```
RuslanLavrov, line 685: Hello @okolawole-ebsco, looks like we don't need to wrap the updateParsedRecord() method code with executeBlocking because the `transaction()` method is asynchronous and does not perform blocking operations

okolawole-ebsco: Good point. I have reverted that method.

PR #464, opened 2022-06-16
MODSOURCE-509 Data Import Updates should add 035 field from 001/003, if it's not HRID or already exists

AivarIUsupov:
## Purpose
When Inventory Single Record Import (or Data Import Update) is used to overlay an existing Instance, the resulting SRS MARC Bib should have correct 035 field (001 should not be HRID of an existing record or 035$a field already exists)

## Approach
Changed HRID manipulation logic for UPDATE actions (moved from SRM to SRS).

#### TODOS and Open Questions
- [ ] Merge only with mod-source-record-manager changes.
- [ ] Change logic for 001 field validation after comments in [MODSOURCE-509](https://issues.folio.org/browse/MODSOURCE-509) added
- [ ] Cover with tests if needed

## Learning
[MODSOURCE-509](https://issues.folio.org/browse/MODSOURCE-509)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
@@ -4,12 +4,12 @@ import static java.lang.String.format;
 import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
 import static org.apache.commons.lang3.StringUtils.isBlank;
-
 import static org.folio.ActionProfile.Action.MODIFY;
 import static org.folio.ActionProfile.Action.UPDATE;
 import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.ACTION_PROFILE;
 import static org.folio.services.util.AdditionalFieldsUtil.HR_ID_FROM_FIELD;
 import static org.folio.services.util.AdditionalFieldsUtil.addControlledFieldToMarcRecord;
+import static org.folio.services.util.AdditionalFieldsUtil.fillHrIdFieldInMarcRecord;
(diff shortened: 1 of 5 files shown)


okolawole-ebsco (approved):

KaterynaSenchenko:
Thank you @Aliaksandr-Fedasiuk, merging it to unblock the other work on matching by 035s

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
```diff
@@ -80,17 +81,27 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload p
...
-          if(isHridFillingNeeded() || isUpdateOption(marcMappingOption)) {
+          if (isHridFillingNeeded()) {
             addControlledFieldToMarcRecord(changedRecord, HR_ID_FROM_FIELD, hrId, true);
             remove003FieldIfNeeded(changedRecord, hrId);
           }
+          if (isUpdateOption(marcMappingOption)) {
+            Record record = Json.decodeValue(recordAsString, Record.class);
+            JsonObject externalEntity = new JsonObject(entityAsString);
+
+            fillHrIdFieldInMarcRecord(Pair.of(record, externalEntity), true);
```
RuslanLavrov, line 103: looks like fillHrIdFieldInMarcRecord() method does nothing because it expects pair of record and instance but here `externalEntity` is a matched marc record. Since a record does not have hrid, hrid [here ](https://github.com/folio-org/mod-source-record-storage/blob/767a2825c8a6bfcdd835e7dd2564d35854f7d1d8/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java#L298) always will be null and therefore any `if` blocks will not be executed in the  fillHrIdFieldInMarcRecord() method.

PR #468, opened 2022-06-21
MODINV-686 - Calculate generation while saving record on instance post-processing

RuslanLavrov:
## Purpose
to calculate generation along with saving incoming record on instance update post-processing if the incoming record has matchedId same as an existing record

## Learning
[MODINV-686](https://issues.folio.org/browse/MODINV-686)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
@@ -33,6 +33,7 @@ import io.vertx.kafka.client.producer.KafkaHeader;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.folio.services.RecordService;
 import org.jooq.Condition;
 
 import org.folio.DataImportEventPayload;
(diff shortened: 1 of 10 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
```diff
@@ -73,13 +74,15 @@ public abstract class AbstractPostProcessingEventHandler implements EventHandler
   private final KafkaConfig kafkaConfig;
   private final MappingParametersSnapshotCache mappingParamsCache;
   private final Vertx vertx;
+  private final RecordService recordService;
 
-  public AbstractPostProcessingEventHandler(final RecordDao recordDao, KafkaConfig kafkaConfig,
+  public AbstractPostProcessingEventHandler(final RecordDao recordDao, KafkaConfig kafkaConfig, RecordService recordService,
```
SerhiiNosko, line 79: To fully replace usage of RecordDao to use only RecordService is not possible, because its necessary big refactoring?

VRohach, line 79: @RuslanLavrov  Could you please describe - does the main reason to change using dao on service here - is to avoid retrieving incoming record by instanceId. Am I right?

Abdulkhakimov (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
RuslanLavrov: I checked the ability to fully replace RecordDao with RecordService, mostly it requires to declare updateParsedRecord() method in the RecordService, made appropriate refactoring

SerhiiNosko: looks great, thanks

SerhiiNosko (approved):

VRohach (approved):

PR #473, opened 2022-07-28
MODSOURCE-509: Improve generation fo 035 field

Aliaksandr-Fedasiuk:
## Purpose
Data Import Updates should add 035 field from 001/003, if it's not HRID or already exists

## Approach
The data from 001 and from 003 fields should not be lost and field 035 should not contain duplicates.
The field 003 should be deleted during import.
For creating MARC BIB records:
- when 035 field isn't present in MARC BIB, this field should be created;

For updating MARC BIB records:
- when 001 field is not equal to hrId then this data should be saved in a new unique 035 field.

## Learning
[MODSOURCE-509](https://issues.folio.org/browse/MODSOURCE-509)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
@@ -314,6 +314,14 @@ public final class AdditionalFieldsUtil {
     }
   }
 
+  public static void fill035FieldInMarcRecord(Record record, String incoming001) {
+    String originalHrIdPrefix = getValueFromControlledField(record, HR_ID_PREFIX_FROM_FIELD);
+    String incoming035 = mergeFieldsFor035(originalHrIdPrefix, incoming001);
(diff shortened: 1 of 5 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -314,6 +314,14 @@ public static void fillHrIdFieldInMarcRecord(Pair<Record, JsonObject> recordInst
     }
   }
 
+  public static void fill035FieldInMarcRecord(Record record, String incoming001) {
```
VRohach, line 317: @Aliaksandr-Fedasiuk  I think there should be better to name this method smth like "fill035FieldInMarcRecordIfNotExists". What do you think?

VRohach (approved):

KaterynaSenchenko (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
Aliaksandr-Fedasiuk: thank you, @VRohach! that’s a good point

RuslanLavrov (approved):

PR #476, opened 2022-08-07
MODSOURCE-528: Improve generation fo 035 field

Aliaksandr-Fedasiuk:
## Purpose
Data Import Updates should add 035 field from 001/003, if it's not HRID or already exists

## Approach
The data from 001 and from 003 fields should not be lost and field 035 should not contain duplicates.
The field 003 should be deleted during import.
For creating MARC BIB records:
- when 035 field isn't present in MARC BIB, this field should be created;

For updating MARC BIB records:
- when 001 field is not equal to hrId then this data should be saved in a new unique 035 field.

## Learning
[MODSOURCE-509](https://issues.folio.org/browse/MODSOURCE-528)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
@@ -257,7 +257,7 @@ public abstract class AbstractPostProcessingEventHandler implements EventHandler
     setExternalIds(externalIdsHolder, externalId, externalHrid);
     boolean isAddedField = addFieldToMarcRecord(record, TAG_999, 'i', externalId);
     if (isHridFillingNeeded()) {
-      fillHrIdFieldInMarcRecord(Pair.of(record, externalEntity), false);
+      fillHrIdFieldInMarcRecord(Pair.of(record, externalEntity));
     }
     if (!isAddedField) {
       throw new PostProcessingException(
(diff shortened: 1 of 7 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -294,21 +294,31 @@ && isNotEmpty(((ControlField) field).getData())
    *
    * @param recordInstancePair pair of related instance and record
```
VRohach, line 295: @Aliaksandr-Fedasiuk  Please add documentation for the brand new param "isUpdateOption".

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
```diff
@@ -257,7 +257,7 @@ private void executeHridManipulation(Record record, JsonObject externalEntity) {
     setExternalIds(externalIdsHolder, externalId, externalHrid);
     boolean isAddedField = addFieldToMarcRecord(record, TAG_999, 'i', externalId);
     if (isHridFillingNeeded()) {
-      fillHrIdFieldInMarcRecord(Pair.of(record, externalEntity));
+      fillHrIdFieldInMarcRecord(Pair.of(record, externalEntity), false);
```
VRohach, line 260: @Aliaksandr-Fedasiuk  If this method always uses with "false" argument - maybe just remove this argument and rename method? What do you think?

VRohach (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
Aliaksandr-Fedasiuk: Ready

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
Aliaksandr-Fedasiuk: Ready. Thank you

okolawole-ebsco (approved):

PR #477, opened 2022-08-12
MODSOURCE-531: Can't update "MARC" record, which was created by stopped import job.

VRohach:
## Purpose
 If import of "MARC" records was stopped by user, the created records impossible to update via "Data import" app because it leads to error. The error message: "io.vertx.pgclient.PgException: ERROR: duplicate key value violates unique constraint "idx_records_matched_id_gen" (23505)" displaying when user click on the name of imported record.

## Approach
Extended statuses via CANCELLED.

JIRA: https://issues.folio.org/browse/MODSOURCE-531


okolawole-ebsco:
any tests possible?

VRohach:
> any tests possible?

Hello @okolawole-ebsco ! I don't think so, that there is an opportunity to add tests for the status.

KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

PR #478, opened 2022-08-14
MODSOURCE-528: Field 035 with hrId should not be created if MARC BIB already exists

Aliaksandr-Fedasiuk:
## Purpose
Field 035 with hrId should not be created if MARC BIB already exists

## Approach
We check if there is a record in the database: if there is, we do not create it and if this extra 035 field already exists, we delete it.

## Learning
[MODSOURCE-528](https://issues.folio.org/browse/MODSOURCE-528)


okolawole-ebsco:
Prior to this change, where was the latest record gotten from the db? I am assuming that prior to this change, SRM add 035 to an existing record and saved to the database. Where is the existing record?
I am asking to see going to the database to check for this field is necessary.

PR #480, opened 2022-08-22
MODSOURCE-528: Remove 035 with actual hrId when exists

Aliaksandr-Fedasiuk:
## Purpose
When SRS receives 035 field that contains actual hrid value this field should be removed before saving in DB.

## Approach
Removing the incorrect 035 field.

## Learning
[MODSOURCE-528](https://issues.folio.org/browse/MODSOURCE-528)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
@@ -152,12 +152,12 @@ public final class AdditionalFieldsUtil {
    * remove field from marc record
    *
    * @param record   record that needs to be updated
-   * @param field    tag of the field
+   * @param fieldName    tag of the field
    * @param subfield subfield of the field
    * @param value    value of the field
    * @return true if succeeded, false otherwise
    */
-  public static boolean removeField(Record record, String field, char subfield, String value) {
+  public static boolean removeField(Record record, String fieldName, char subfield, String value) {
     boolean isFieldRemoveSucceed = false;
     try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
       if (record != null && record.getParsedRecord() != null && record.getParsedRecord().getContent() != null) {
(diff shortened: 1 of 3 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -152,34 +151,79 @@ public static boolean addControlledFieldToMarcRecord(Record record, String field
...
-          VariableField variableField = marcRecord.getVariableField(field);
-          if (variableField != null) {
-            marcRecord.removeVariableField(variableField);
+          List<VariableField> variableFields = marcRecord.getVariableFields(field);
+          for (VariableField variableField : variableFields) {
+            if (StringUtils.isEmpty(value)) {
+              marcRecord.removeVariableField(variableField);
+              isFieldRemoveSucceed = true;
+              break;
+            } else {
```
VRohach, line 175: @Aliaksandr-Fedasiuk  I think we can use here "else-if" instead if inside else.

VRohach (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -152,34 +151,79 @@ public static boolean addControlledFieldToMarcRecord(Record record, String field
...
+              if (isFieldContainsValue(variableField, subfield, value)) {
```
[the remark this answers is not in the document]
Aliaksandr-Fedasiuk: Thank you, @okolawole-ebsco, completed

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
Aliaksandr-Fedasiuk: Thank you, @VRohach , completed

okolawole-ebsco (approved):

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #486, opened 2022-10-03
[MODSOURCE-507]Reduce Computation of MARC4J Records in AdditionalFieldsUtil

KyryloShulzhenko:
Continuation of similar work done in [MODSOURMAN-791](https://issues.folio.org/browse/MODSOURMAN-791). AdditionalFieldsUtil in SRS should cache parsed content to reduce CPU utilization.

Acceptance Criteria
AdditionalFieldsUtil should not process a parsed content string that it has seen before.
Cache should have the "WeakKeys" property so that items in the cache can be garbage collected.
Maximum size of items in cache should be set to a suitable amount.
https://issues.folio.org/browse/MODSOURCE-507

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
@@ -1,5 +1,11 @@
 package org.folio.services.util;
 
+import com.github.benmanes.caffeine.cache.CacheLoader;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.github.benmanes.caffeine.cache.stats.CacheStats;
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
 import io.vertx.core.json.JsonObject;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
@@ -15,11 +21,7 @@ import org.marc4j.MarcJsonWriter;
 import org.marc4j.MarcReader;
 import org.marc4j.MarcStreamWriter;
 import org.marc4j.MarcWriter;
-import org.marc4j.marc.ControlField;
-import org.marc4j.marc.DataField;
-import org.marc4j.marc.MarcFactory;
-import org.marc4j.marc.Subfield;
-import org.marc4j.marc.VariableField;
+import org.marc4j.marc.*;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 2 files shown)


RomanChernetskyi (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -15,11 +21,7 @@
 import org.marc4j.MarcReader;
 import org.marc4j.MarcStreamWriter;
 import org.marc4j.MarcWriter;
-import org.marc4j.marc.ControlField;
-import org.marc4j.marc.DataField;
-import org.marc4j.marc.MarcFactory;
-import org.marc4j.marc.Subfield;
-import org.marc4j.marc.VariableField;
+import org.marc4j.marc.*;
```
Aliaksandr-Fedasiuk, line 24: @KyryloShulzhenko, could you please remove "*"? Just set a huge value in imports in intellijidea.

Aliaksandr-Fedasiuk (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -509,4 +570,22 @@ private static boolean isNotProtected(List<MarcFieldProtectionSetting> fieldProt
...
+      && record.getParsedRecord() != null
+      && !org.apache.commons.lang3.StringUtils.isBlank(record.getParsedRecord().getContent().toString())) {
+      try {
+        return parsedRecordContentCache.get(record.getParsedRecord().getContent());
+      } catch (Exception e) {
+        LOGGER.error("Error during the transformation to marc record", e);
+        MarcReader reader = buildMarcReader(record);
+        if (reader.hasNext()) {
+          return reader.next();
+        }
```
RuslanLavrov, line 585: Hello @KyryloShulzhenko, do we need here to try to parse a record one more time in case of exception? Because it looks like this "catch" block can be called due to record parsing error in the `parsedRecordContentCacheLoader`
block, while calling parsedRecordContentCache.get() method of the cache.

KyryloShulzhenko: Hello @RuslanLavrov  parsedRecordContentCache.get() in some cases don`t parse record and throw exception. This code help to parse record

RuslanLavrov: in such case, propose to set the "warn" logging level here at the "catch" block, since the method can return a successfully parsed record

PR #489, opened 2022-10-19
MODSOURCE-552 Extend GET /source-storage/source-records/<Id> to fetch deleted record

Shans-Kaluhin:
## Approach
- add `state` query parameter
- add tests cases

## Learning
https://issues.folio.org/browse/MODSOURCE-552

### DIFF
--- a/mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SourceRecordApiTest.java
+++ b/mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SourceRecordApiTest.java
@@ -357,6 +357,21 @@ public class SourceRecordApiTest extends AbstractRestVerticleTest {
   }
 
   @Test
+  public void shouldReturnSpecificMarcBibSourceRecordOnGetByRecordExternalIdAndRecordState(TestContext testContext) {
+    returnSpecificMarcSourceRecordOnGetByRecordExternalIdAndRecordState(testContext, snapshot_2, RecordType.MARC_BIB);
+  }
+
+  @Test
+  public void shouldReturnSpecificMarcAuthoritySourceRecordOnGetByRecordExternalIdAndRecordState(TestContext testContext) {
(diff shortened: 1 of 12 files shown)


mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SourceRecordApiTest.java
```diff
@@ -1661,6 +1676,51 @@ private void returnSpecificMarcSourceRecordOnGetByRecordExternalId(TestContext t
     async.complete();
   }
 
+  private void returnSpecificMarcSourceRecordOnGetByRecordExternalIdAndRecordState(TestContext testContext,
```
sherzodn, line 1679: can we use Parameterized test here?

mod-source-record-storage-server/src/test/java/org/folio/rest/util/QueryParamUtilTest.java
```diff
@@ -75,4 +76,34 @@ public void shouldThrowBadRequestExceptionForUnknownRecordType() {
     QueryParamUtil.toRecordType("UNKNOWN");
   }
 
+  @Test
```
sherzodn, line 79: maybe you can do it with @Parameterized test

psmagin: There could be a problem because this module is using junit4, which as I remember doesn't have an easy solution for parameterized tests.

mod-source-record-storage-server/src/test/java/org/folio/rest/impl/SourceRecordApiTest.java
Shans-Kaluhin: Good idea, but it requires to add new dependency. And all tests in this module are written that way.
So our test will be different or we have to rewrite all tests

sherzodn (approved):

psmagin (approved):

PR #511, opened 2023-01-18
MODSOURCE-562 - Handle bib-authority link when user updates a bib record via data import

sherzodn:
## Purpose
_Handle bib-authority link when user updates a bib record via data import._

## Approach
On modifying part of bib records update flow do:

- Check old and new records to identify if there are changes to fields with $9 subfield
- If yes then fetch from mod-entities-links records with current linkage for current instance_id
- Provide links data to MarcRecordModifier that is part of data-import-processing-core
- MarcRecordsModifier will update links data, after that make a call to update links in mod-entites-links

#### TODOS and Open Questions
- [ ] Check logging.
## Learning
https://issues.folio.org/browse/MODSOURCE-562


sherzodn:
image
Tests are passing successfully in local run

hasanepam (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
```diff
@@ -49,13 +53,14 @@ public abstract class AbstractUpdateModifyEventHandler implements EventHandler {
...
   private static final String PAYLOAD_HAS_NO_DATA_MSG =
     "Failed to handle event payload, cause event payload context does not contain required data to modify MARC record";
-  private static final String MAPPING_PARAMETERS_NOT_FOUND_MSG =
-    "MappingParameters snapshot was not found by jobExecutionId '%s'";
+  private static final String MAPPING_PARAMETERS_NOT_FOUND_MSG = "MappingParameters snapshot was not found by jobExecutionId '%s'";
+
   protected RecordService recordService;
   protected MappingParametersSnapshotCache mappingParametersCache;
   protected Vertx vertx;
+  protected InstanceLinkClient instanceLinkClient;
```
psmagin, line 61: I think it could be moved to bib class.

descriptors/ModuleDescriptor-template.json
```diff
@@ -226,6 +226,30 @@
         }
       ]
     },
+    {
```
viacheslavkol, line 229: srs does not provide this endpoint.
It must be added to "requires" section (check mod-search for reference)

viacheslavkol (changes requested):

descriptors/ModuleDescriptor-template.json
sherzodn: done

psmagin (approved):

viacheslavkol (approved):

PR #512, opened 2023-01-18
MODSOURCE-567 Link update report: send event if update failed

Shans-Kaluhin:
When SRS consumes an event from `links.instance-authority` it tries to update related SRS records. If the update failed then send an event to `links.instance-authority-stats` topic.

https://issues.folio.org/browse/MODSOURCE-567

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityLinkChunkKafkaHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityLinkChunkKafkaHandler.java
@@ -55,7 +55,7 @@ import org.springframework.stereotype.Component;
 @Component
 public class AuthorityLinkChunkKafkaHandler implements AsyncRecordHandler<String, String> {
 
-  public static final String SRS_BIB_UPDATE_TOPIC = MARC_BIB.moduleName() + "." + MARC_BIB.topicName();
+  public static final String SRS_BIB_UPDATE_TOPIC = MARC_BIB.moduleTopicName();
 
   private static final char AUTHORITY_ID_SUBFIELD = '9';
   private static final AtomicLong INDEXER = new AtomicLong();
@@ -222,6 +222,7 @@ public class AuthorityLinkChunkKafkaHandler implements AsyncRecordHandler<String
           + " Total number of records: {}, successful: {}, failures: {}",
         event.getJobId(), event.getAuthorityId(),
         batchResponse.getTotalRecords(), batchResponse.getRecords().size(), errors);
+      // send report
     }
 
     return toMarcBibUpdateEvents(batchResponse, event);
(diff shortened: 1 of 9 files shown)


hasanepam (approved):

Shans-Kaluhin:
This PR waits for merging this PR's:
`ramls`: https://github.com/folio-org/data-import-raml-storage/pull/271
`folio-kafka-wrapper`: https://github.com/folio-org/folio-kafka-wrapper/pull/35

hasanepam (approved):

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityLinkChunkKafkaHandler.java
```diff
@@ -262,49 +293,58 @@ private Future<BibAuthorityLinksUpdate> createSnapshot(BibAuthorityLinksUpdate b
...
       return Future.failedFuture(e);
     }
-    return promise.future();
+    return promise.future().onFailure(th -> LOGGER.error("Failed to send {} event for jobId {}.", topic.topicName(), jobId, th));
   }
 
-  private KafkaProducerRecord<String, String> createKafkaProducerRecord(MarcBibUpdate marcBibUpdate,
+  private KafkaProducerRecord<String, String> createKafkaProducerRecord(KafkaTopic topic, String tenant, Object record,
                                                                         List<KafkaHeader> kafkaHeaders) {
-    var topicName = createTopicNameNoNamespace(SRS_BIB_UPDATE_TOPIC, marcBibUpdate.getTenant(), kafkaConfig);
```
viacheslavkol, line 305: This method could be also removed from utils

mod-source-record-storage-server/src/test/java/org/folio/services/AuthorityLinkChunkKafkaHandlerTest.java
```diff
@@ -63,32 +67,34 @@ public class AuthorityLinkChunkKafkaHandlerTest extends AbstractLBServiceTest {
   private static final String PARSED_MARC_RECORD_UNLINKED = "src/test/resources/parsedMarcRecordUnlinked.json";
   private static final String KAFKA_KEY_NAME = "test-key";
   private static final String KAFKA_TEST_HEADER = "test";
-  private static final String KAFKA_CONSUMER_TOPIC = getTopicName(AUTHORITY_INSTANCE_LINKS_TOPIC);
-  private static final String KAFKA_PRODUCER_TOPIC = getTopicName(SRS_BIB_UPDATE_TOPIC);
+  private static final String KAFKA_CONSUMER_TOPIC = getTopicName(INSTANCE_AUTHORITY);
+  private static final String KAFKA_KAFKA_SRS_BIB_PRODUCER_TOPIC = getTopicName(MARC_BIB);
```
viacheslavkol, line 71: Naming trouble

viacheslavkol (approved):

psmagin (changes requested):
I don't think that we should send success events from srs. It's like a completion event for the whole update, and it can be completed fully only in inventory.

PR #535, opened 2023-03-30
MODSOURCE-618 - Added DB_CONNECTION_TIMEOUT env variable to make pull timeout configurable

RuslanLavrov:
## Purpose
to make connection pull timeout configurable

## Learning
[MODSOURCE-618](https://issues.folio.org/browse/MODSOURCE-618)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
@@ -45,6 +45,8 @@ public class PostgresClientFactory {
   public static final String PASSWORD = "password";
   public static final String USERNAME = "username";
   public static final String DB_MAXPOOLSIZE = "maxPoolSize";
+  private static final String CONNECTION_TIMEOUT = "DB_CONNECTION_TIMEOUT";
+  private static final String DEFAULT_CONNECTION_TIMEOUT_VALUE = "30";
   private static final String IDLE_TIMEOUT = "connectionReleaseDelay";
   private static final String MODULE_NAME = ModuleName.getModuleName();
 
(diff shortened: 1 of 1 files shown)


VRohach (approved):

KaterynaSenchenko (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
```diff
@@ -169,9 +171,13 @@ private static PgPool getCachedPool(Vertx vertx, String tenantId) {
     }
 
     Integer maxPoolSize = postgresConfig.getInteger(DB_MAXPOOLSIZE, DB_MAXPOOLSIZE_DEFAULT_VALUE);
+    int connectionTimeout = Integer.parseInt(System.getenv().getOrDefault(CONNECTION_TIMEOUT, DEFAULT_CONNECTION_TIMEOUT_VALUE));
     LOG.info("getCachedPool:: Creating new database connection for tenant {} with poolSize {}", tenantId, maxPoolSize);
     PgConnectOptions connectOptions = getConnectOptions(tenantId);
-    PoolOptions poolOptions = new PoolOptions().setMaxSize(maxPoolSize);
+    PoolOptions poolOptions = new PoolOptions()
+      .setConnectionTimeout(connectionTimeout)
+      .setIdleTimeoutUnit(TimeUnit.SECONDS)
```
okolawole-ebsco, line 179: I think you meant `setConnectionTimeoutUnit` here instead of `setIdleTimeoutUnit`. The default is still "seconds" for `setConnectionTimeoutUnit` so the intent is still there. But we should change to accurately depict what we mean.

RuslanLavrov: thank you for spotting this, I will correct it

PR #540, opened 2023-05-03
MODSOURCE-628 Create bulk endpoint to fetch parsed records

Shans-Kaluhin:
## Approach
feat(authority-control): Create bulk endpoint to fetch parsed records

## Learning
https://issues.folio.org/browse/MODSOURCE-628

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -257,6 +257,10 @@
     },
     {
       "id": "instance-authority-links",
+      "version": "2.0"
+    },
+    {
+      "id": "instance-authority-linking-rules",
       "version": "1.0"
     }
   ],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -219,6 +219,15 @@
           "methods": [
             "POST"
           ],
+          "pathPattern": "/source-storage/batch/parsed-records/fetch",
+          "permissionsRequired": [
+            "source-storage.records.fetch"
+          ]
+        },
+        {
+          "methods": [
+            "POST"
+          ],
           "pathPattern": "/source-storage/batch/verified-records",
           "permissionsRequired": [
             "source-storage.verified.records"
@@ -306,6 +315,11 @@
       "description": "Put Record"
     },
     {
+      "permissionName": "source-storage.records.fetch",
+      "displayName": "Source Storage - fetch record",
+      "description": "Fetch Record"
+    },
+    {
       "permissionName": "source-storage.records.update",
       "displayName": "Source Storage - update record",
       "description": "Update Record's fields"
@@ -338,6 +352,7 @@
         "source-storage.records.get",
         "source-storage.records.post",
         "source-storage.records.put",
+        "source-storage.records.fetch",
         "source-storage.records.delete",
         "source-storage.records.update",
         "source-storage.sourceRecords.get",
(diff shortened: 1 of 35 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -215,6 +215,15 @@
...
         },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/source-storage/batch/parsed-records/fetch",
+          "permissionsRequired": [
+            "source-storage.records.fetch"
+          ]
+        },
```
psmagin, line 226: Please increase minor interface version

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageBatchImpl.java
```diff
@@ -101,4 +102,30 @@ public void putSourceStorageBatchParsedRecords(RecordCollection entity, Map<Stri
...
+                                                       Context vertxContext) {
+    vertxContext.runOnContext(v -> {
+      try {
+        recordService.fetchParsedRecords(entity, tenantId)
+          .map(parsedRecordsBatchResponse -> {
+            if (!parsedRecordsBatchResponse.getRecords().isEmpty()) {
+              return PostSourceStorageBatchParsedRecordsFetchResponse.respond200WithApplicationJson(parsedRecordsBatchResponse);
+            } else {
+              LOG.warn("postSourceStorageBatchParsedRecordsFetch:: Batch of parsed records was fetched, but no records were found");
+              return PostSourceStorageBatchParsedRecordsFetchResponse.respond500WithApplicationJson(parsedRecordsBatchResponse);
```
psmagin, line 119: I think it could be 400 or 404. 500 is for unexpected cases.

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -170,6 +175,22 @@ public Future<ParsedRecordsBatchResponse> updateParsedRecords(RecordCollection r
...
+    var idType = IdType.valueOf(fetchRequest.getConditions().getIdType());
+    if (ids.isEmpty()) {
+      Promise<RecordCollection> promise = Promise.promise();
+      promise.complete(new RecordCollection().withTotalRecords(0));
+      return promise.future();
+    }
+
+    var idsCondition = getExternalIdsCondition(ids, idType);
+    var recordType = toRecordType(fetchRequest.getRecordType().name());
+    return recordDao.getRecords(idsCondition, recordType, emptyList(), 0, ids.size(), tenantId)
```
psmagin, line 190: I believe that this method in `recordDao` do to many operations, and we need the endpoint to be as performance as possible.
There is a lot of stuff that clould be excluded from the response. Like: metadata, additionalInfo, errorRecord, rawRecord and so on. We need just parsed content and ids.

psmagin (changes requested):

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageBatchImpl.java
Shans-Kaluhin: In most cases SRS returns 500 status with empty list of records.
But I believe it's better to return 200 with empty list

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
Shans-Kaluhin: Done. Added a new Dto

PR #550, opened 2023-06-16
MODSOURCE-636 - Implement async migration service

RuslanLavrov:
## Purpose
Migration script to populate marc_records_tracking table and version for marc_indexers table should be run asynchronously (not at the module deployment time) through REST API call. 
It's necessary to provide API for the management of asynchronous migrations to run migration scripts, which take a long time to complete, manually and separately from the module deployment time.

## Approach
* implement asynchronous endpoint `POST /source-storage/migrations/jobs` to initiate asynchronous migration job and appropriate scripts run.
  The endpoint allows only one migration job to be in progress. If during the endpoint call another migration job is still being IN_PROGRESS, then the endpoint will respond 409 Conflict in response.
* implement endpoint `GET /source-storage/migrations/jobs/{jobId}` to check the status of migration job execution by its id.
* move scripts to populate "marc_records_tracking" table and to update version for "marc_indexers" table to async-migration scripts (`/async_migration` folder) to be run through API.
* add tests

## Learning
[MODSOURCE-636](https://issues.folio.org/browse/MODSOURCE-636)

Requires: PR https://github.com/folio-org/data-import-raml-storage/pull/285

The PR build will be fixed after https://github.com/folio-org/data-import-raml-storage/pull/285 PR is merged

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/AsyncMigrationJobDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/AsyncMigrationJobDaoImpl.java
@@ -37,12 +37,12 @@ public class AsyncMigrationJobDaoImpl implements AsyncMigrationJobDao {
   }
 
   @Override
-  public Future<String> save(AsyncMigrationJob migrationJob, String tenantId) {
+  public Future<Void> save(AsyncMigrationJob migrationJob, String tenantId) {
     LOG.trace("save:: Saving async migration job with id {} for tenant {}", migrationJob.getId(), tenantId);
     return getQueryExecutor(tenantId).executeAny(dslContext -> dslContext
         .insertInto(ASYNC_MIGRATION_JOBS)
         .set(mapToDatabaseRecord(migrationJob)))
-      .map(migrationJob.getId());
+      .mapEmpty();
   }
 
   @Override
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/AsyncMigrationJobDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/AsyncMigrationJobDaoImpl.java
@@ -24,7 +24,7 @@ import static java.lang.String.format;
 import static org.folio.rest.jooq.Tables.ASYNC_MIGRATION_JOBS;
 
 @Repository
-public class MigrationJobDaoImpl implements MigrationJobDao {
+public class AsyncMigrationJobDaoImpl implements AsyncMigrationJobDao {
 
   private static final Logger LOG = LogManager.getLogger();
   public static final String JOB_NOT_FOUND_MSG = "Async migration job was not found by id: '%s'";
@@ -32,7 +32,7 @@ public class MigrationJobDaoImpl implements MigrationJobDao {
   private PostgresClientFactory postgresClientFactory;
 
   @Autowired
-  public MigrationJobDaoImpl(PostgresClientFactory postgresClientFactory) {
+  public AsyncMigrationJobDaoImpl(PostgresClientFactory postgresClientFactory) {
     this.postgresClientFactory = postgresClientFactory;
   }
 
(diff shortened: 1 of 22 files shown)


okolawole-ebsco:
Should customers be allowed to perform their duties if the migration is still ongoing?

mod-source-record-storage-server/src/main/java/org/folio/dao/AsyncMigrationJobDaoImpl.java
```diff
@@ -0,0 +1,123 @@
...
+      asyncMigrationJobsDbRecord.setError(migrationJob.getErrorMessage());
+    }
+    return asyncMigrationJobsDbRecord;
+  }
+
+  private AsyncMigrationJob mapRowToAsyncMigrationJob(Row row) {
+    AsyncMigrationJobs pojo = RowMappers.getAsyncMigrationJobsMapper().apply(row);
+    AsyncMigrationJob asyncMigrationJob = new AsyncMigrationJob()
+      .withId(pojo.getId().toString())
+//      .withMigrations(Arrays.asList(pojo.getMigrations()))
```
okolawole-ebsco, line 109: remove comment?

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageMigrationsJobsImpl.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+
+import javax.ws.rs.core.Response;
+import java.util.Map;
+
+import static java.lang.String.format;
+
+public class SourceStorageMigrationsJobsImpl implements SourceStorageMigrationsJobs {
+
+  public static final String NOT_FOUND_MSG = "Async migration job with id '%s' was not found";
```
okolawole-ebsco, line 21: Does this need to be public?

RuslanLavrov: no, just haven't changed this modifier after extraction value to constants by IDE

RuslanLavrov:
> Should customers be allowed to perform their duties if the migration is still ongoing?

It is recommended to perform migration during the period of minimal activity, such as nighttime hours, for example. 
But, from the discussion with backenders it was proposed to accommodate a case when async migration can be executed slightly later after the module upgrade. As a result, some marc records can be imported after the module upgrade but before the asynchronous migration execution. To handle such case a couple of statements were changed in the script to prevent overwriting the "marc_indexers"."version" to 0 during migration execution for records imported after module upgrade. 
And to avoid redundant records insert into the "marc_records_tracking" table for those imported marc records during migration execution.

mod-source-record-storage-server/src/main/java/org/folio/dao/AsyncMigrationJobDaoImpl.java
RuslanLavrov: removed

Aliaksandr-Fedasiuk (approved):

mod-source-record-storage-server/src/main/java/org/folio/rest/impl/SourceStorageMigrationsJobsImpl.java
```diff
@@ -0,0 +1,54 @@
...
+  @Autowired
+  private AsyncMigrationJobService asyncMigrationJobService;
+
+  public SourceStorageMigrationsJobsImpl(Vertx vertx, String tenantId) { //NOSONAR
+    SpringContextUtil.autowireDependencies(this, Vertx.currentContext());
+    this.tenantId = tenantId;
+  }
+
+  @Override
+  public void postSourceStorageMigrationsJobs(AsyncMigrationJobInitRq entity, Map<String, String> okapiHeaders,
```
okolawole-ebsco, line 34: Does this mean any user can initiate a migration?

RuslanLavrov: only users which have "source-storage.migrations.post" permission can initiate migration

PR #551, opened 2023-06-20
MODSOURCE-639: Update Leader position when authority record is deleted

mukhiddin-yusuf:
Update Leader position when authority record is deleted

## Purpose
_Update Leader position when authority record is deleted_

## Approach
_When user deletes an authority record update Parsed Record Leader position 05 to value of 'd'_

## Closes
[_MODSOURCE-639_](https//issues.folio.org/browse/MODSOURCE-639)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
@@ -6,6 +6,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.ActionProfile;
 import org.folio.DataImportEventPayload;
+import org.folio.dao.util.RecordType;
 import org.folio.processing.events.services.handler.EventHandler;
 import org.folio.processing.exceptions.EventProcessingException;
 import org.folio.rest.jaxrs.model.ExternalIdsHolder;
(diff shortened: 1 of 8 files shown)


hasanepam (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
```diff
@@ -67,11 +68,18 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload p
   private void handlePayload(DataImportEventPayload payload, CompletableFuture<DataImportEventPayload> future) {
     var record = Json.decodeValue(payload.getContext().get(getRecordKey()), Record.class);
     LOG.info("handlePayload:: Handling 'delete' event for the record id = {}", record.getId());
-    recordService.updateRecordsState(record.getMatchedId(), RecordState.DELETED, payload.getTenant())
+    ParsedRecordDaoUtil.updateLeaderStatus(record.getParsedRecord(), 'd');
+    record.setState(Record.State.DELETED);
+    recordService.getRecordById(record.getId(), payload.getTenant())
+        .compose(optionalRecord -> optionalRecord
+          .map(existingRecord -> recordService.saveRecord(record, payload.getTenant()))
+          .orElse(Future.succeededFuture(null)))
```
psmagin, line 76: `updateRecordsState` method differs in behavior with `saveRecord` method.
`updateRecordsState` method set status to all generations of the records, instead of `saveRecord` that will create a new generation of a record
image

psmagin (changes requested):

psmagin (approved):

PR #559, opened 2023-08-03
MODSOURCE-676 Implement retries for queries with specific exceptions

okolawole-ebsco:
## Purpose
Retry database queries that fail intermittently with non-actionable exceptions

## Approach
ByteBuddy library is used to generate bytecode that will append `.recover` from Vert.x to the future representing the result of the database query. The retry logic is embedded within the `.recover` composition of the database future. Retry logic is only performed for the `query` and `transaction` methods of ReactiveClassicGenericQueryExecutor from JOOQ-Vertx.
Retry will only occur for two exceptions as of this PR: `io.vertx.core.impl.NoStackTraceThrowable` and `io.netty.channel.StacklessClosedChannelException`

## Learning
https://issues.folio.org/browse/MODSOURCE-676

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/QueryExecutorInterceptor.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/QueryExecutorInterceptor.java
@@ -1,0 +1,146 @@
+package org.folio.dao;
+
+import io.github.jklingsporn.vertx.jooq.classic.reactivepg.ReactiveClassicGenericQueryExecutor;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+import io.vertx.core.impl.NoStackTraceThrowable;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.implementation.MethodDelegation;
+import net.bytebuddy.implementation.SuperMethodCall;
+import net.bytebuddy.matcher.ElementMatchers;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.nio.channels.ClosedChannelException;
+import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+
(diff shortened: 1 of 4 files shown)


KaterynaSenchenko (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/QueryExecutorInterceptor.java
```diff
@@ -0,0 +1,148 @@
...
+  @SuppressWarnings("unused")
+  public static <U> Future<U> query(
+    @net.bytebuddy.implementation.bind.annotation.SuperCall Callable<Future<U>> superCall
+  ) {
+    LOGGER.trace("query method of ReactiveClassicGenericQueryExecutor proxied");
+    return retryOf(() -> {
+      try {
+        return superCall.call();
+      } catch (Throwable e) {
+        LOGGER.error("Something happened while attempting to make proxied call for query method", e);
```
VRohach, line 87: @okolawole-ebsco As far as I remember for our current approach we set all errors as WARN-level for better errors-handling process on the envs. Or do you think this is an exceptional case for the whole module here? (and other places in this PR)

VRohach (approved):

Aliaksandr-Fedasiuk (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/QueryExecutorInterceptor.java
okolawole-ebsco: Yeah I think it is an exceptional case. I don't expect this to happen often

yaroslav-epam (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/QueryExecutorInterceptor.java
VRohach: @okolawole-ebsco  Thank you!

PR #563, opened 2023-08-22
MODSOURCE-674 - Ensure only one background job can be triggered to clean up outdated marc indexers

RuslanLavrov:
## Purpose
to prevent module instances from blocking each other while attempting to delete marc indexers on the same tenant

## Approach
* apply advisory lock for deletion marc indexers on certain tenant to prevent this deletion from being executed by multiple module instances for the same tenant at the same time

## Learning
[MODSOURCE-674](https://issues.folio.org/browse/MODSOURCE-674)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
@@ -96,6 +96,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static java.lang.String.format;
+import static org.folio.dao.util.AdvisoryLockUtil.acquireLock;
 import static org.folio.dao.util.ErrorRecordDaoUtil.ERROR_RECORD_CONTENT;
 import static org.folio.dao.util.ParsedRecordDaoUtil.PARSED_RECORD_CONTENT;
 import static org.folio.dao.util.RawRecordDaoUtil.RAW_RECORD_CONTENT;
@@ -125,7 +126,6 @@ import static org.jooq.impl.DSL.primaryKey;
 import static org.jooq.impl.DSL.select;
 import static org.jooq.impl.DSL.table;
 import static org.jooq.impl.DSL.trueCondition;
-import static org.jooq.impl.DSL.val;
 
 @Component
 public class RecordDaoImpl implements RecordDao {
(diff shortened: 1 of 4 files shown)


mod-source-record-storage-server/src/test/java/org/folio/dao/RecordDaoImplTest.java
```diff
@@ -138,6 +138,20 @@ public void shouldReturnIdOnStreamMarcRecordIdsWhenThereIsNoTrackingRecordAndFal
     });
   }
 
+  @Test
+  public void shouldReturnFalseWhenPreviousIndexersDeletionIsInProgress(TestContext context) {
+    Async async = context.async();
+
+    recordDao.deleteMarcIndexersOldVersions(TENANT_ID);
```
okolawole-ebsco, line 145: Is there any guarantee that this call would still be running when the next call happens? Thinking it might get flaky.

okolawole-ebsco (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -1127,31 +1129,45 @@ public Future<Void> deleteRecords(int lastUpdatedDays, int limit, String tenantI
...
+          LOG.info("deleteMarcIndexersOldVersions:: Previous marc_indexers old version deletion still ongoing, tenantId: '{}'", tenantId);
+          return Future.succeededFuture(false);
+        }
+        return deleteMarcIndexersOldVersions(txQE, tenantId);
+      }), tenantId);
+  }
+
+  private Future<Boolean> acquireLock(ReactiveClassicGenericQueryExecutor txQE, String tenantId) {
+    return txQE.findOneRow(dsl -> dsl.select(DSL.field("{0}", Boolean.class,
+      DSL.function(GET_LOCK_FUNCTION, SQLDataType.BOOLEAN, val(tenantId.hashCode()))))
```
okolawole-ebsco, line 1144: Thinking maybe we should have a hashcode on a prefix like "delete_marc_indexers_"+tenantId instead of just tenant id. Just so to avoid collisions with actors that may just use the tenant id as the lock id.

okolawole-ebsco:
@RuslanLavrov when is the lock released after deletion?

VRohach:
@RuslanLavrov  Update NEWS please.

VRohach (approved):

RuslanLavrov:
@Olamide, the lock is released on the completion of the transaction for indexers deletion (transaction within which a lock was obtained).

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
RuslanLavrov: added prefix usage for lock obtaining, just initially was thinking of adding prefix/additional lock id on demand when the same approach will be used

mod-source-record-storage-server/src/test/java/org/folio/dao/RecordDaoImplTest.java
RuslanLavrov: there is no strict and explicit mechanism in the test that guarantees it. This behavior is achieved due to the relative "heaviness" of the operations to the DB, which are performed in this method, so, suppose there is a probability of flaky test behavior. Changed the test to avoid the probability of the test getting flaky

PR #565, opened 2023-08-25
[MODSOURCE-530] Send DuplicateException if incoming file contains duplicate records

RomanChernetskyi:
## Purpose
Duplicate records in incoming file causes problems after overlay process with no error reported

## Approach
Changed a way of assigning matched id, now it generated during save record process, matched id is retrieved from 999$s field, in case if the parsed record does not contain this field, assigning id of exiting source record with a same external id

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
@@ -9,13 +9,18 @@ import static org.folio.dao.util.RecordDaoUtil.ensureRecordHasSuppressDiscovery;
 import static org.folio.dao.util.SnapshotDaoUtil.SNAPSHOT_NOT_FOUND_TEMPLATE;
 import static org.folio.dao.util.SnapshotDaoUtil.SNAPSHOT_NOT_STARTED_MESSAGE_TEMPLATE;
 import static org.folio.rest.util.QueryParamUtil.toRecordType;
+import static org.folio.services.util.AdditionalFieldsUtil.TAG_999;
+import static org.folio.services.util.AdditionalFieldsUtil.getFieldFromMarcRecord;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
+import java.util.EnumMap;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import javax.ws.rs.BadRequestException;
 import javax.ws.rs.NotFoundException;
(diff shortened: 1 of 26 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -250,6 +275,50 @@ public Future<Void> updateRecordsState(String matchedId, RecordState state, Reco
     return recordDao.updateRecordsState(matchedId, state, recordType, tenantId);
   }
 
+  private Future<Record> setMatchedIdForRecord(Record record, String tenantId) {
```
okolawole-ebsco, line 278: please add some comments to the body of this method.

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
```diff
@@ -587,10 +599,12 @@ && isNotBlank(record.getParsedRecord().getContent().toString())) {
...
-        MarcReader reader = buildMarcReader(record);
-        if (reader.hasNext()) {
-          return reader.next();
-        }
+        try {
+          MarcReader reader = buildMarcReader(record);
+          if (reader.hasNext()) {
+            return reader.next();
+          }
+        } catch (Exception ignored) {}
```
okolawole-ebsco, line 607: why doesn't it make sense to log the exception?

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
RomanChernetskyi: done, thanks

mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java
RomanChernetskyi: done, thanks

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -250,6 +275,54 @@ public Future<Void> updateRecordsState(String matchedId, RecordState state, Reco
...
+          } else {
+            LOG.warn("setMatchedIdForRecord:: Error while retrieving source record");
+            promise.fail(ar.cause());
+          }
+        });
+    } else {
+      // Set matched id same as record id
+      promise.complete(record.withMatchedId(record.getId()));
+    }
+    return promise.future().onSuccess(r -> addFieldToMarcRecord(r, TAG_999, SUBFIELD_S, r.getMatchedId()));
```
VRohach, line 314: @RomanChernetskyi  Is there an opportunity to divide this method via other methods? The body in this "if" condition seems so complex. Thank you in advance.

VRohach (commented):
@RomanChernetskyi  Update NEWS please.

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
RomanChernetskyi: Done, thanks

VRohach (approved):

yaroslav-epam (approved):

PR #568, opened 2023-09-12
[MODSOURCE-646] Perform MARC To MARC Matching in Local Tenant & Central Tenant

RomanChernetskyi:
## Purpose
When performing MARC to MARC matching, matching should occur in two tenants; central & local to realize ECS. This is because a Shadow Instance will not have a MARC reference in the local tenant. The Shared Instance might have a MARC reference if it has source=MARC.
## Approach
Added cache for centralTenantId and consortiumId
Created logic that performs matching at central tenant if local tenant i in ECS, combines result of both matching

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/caches/ConsortiumConfigurationCache.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/caches/ConsortiumConfigurationCache.java
@@ -21,6 +21,9 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
+/**
+ * Cache for {@link ConsortiumConfiguration}
+ */
 @Component
 public class ConsortiumConfigurationCache {
   private static final Logger LOGGER = LogManager.getLogger();
@@ -38,6 +41,12 @@ public class ConsortiumConfigurationCache {
       .buildAsync();
   }
 
+  /**
+   * Searches for {@link ConsortiumConfiguration} cache by tenant id
+   *
+   * @param params okapi connection parameters
+   * @return future with optional {@link ConsortiumConfiguration}
+   */
   public Future<Optional<ConsortiumConfiguration>> get(OkapiConnectionParams params) {
     try {
       return Future.fromCompletionStage(cache.get(params.getTenantId(), (key, executor) -> loadConsortiumConfiguration(params)));
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/caches/ConsortiumConfigurationCache.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/caches/ConsortiumConfigurationCache.java
@@ -1,0 +1,75 @@
+package org.folio.services.caches;
+
+import com.github.benmanes.caffeine.cache.AsyncCache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.json.JsonArray;
+import org.apache.http.HttpStatus;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.folio.dataimport.util.OkapiConnectionParams;
+import org.folio.dataimport.util.RestUtil;
+import org.folio.services.entities.ConsortiumConfiguration;
+import org.folio.services.exceptions.CacheLoadingException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class ConsortiumConfigurationCache {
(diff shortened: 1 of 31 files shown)


yaroslav-epam (approved):

Maksat-Galymzhan (approved):

KaterynaSenchenko (approved):

RuslanLavrov (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/caches/ConsortiumConfigurationCache.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+@Component
+public class ConsortiumConfigurationCache {
```
VRohach, line 25: @RomanChernetskyi  Let's add some documentation for the new classes.

mod-source-record-storage-server/src/main/java/org/folio/services/entities/ConsortiumConfiguration.java
```diff
@@ -0,0 +1,27 @@
+package org.folio.services.entities;
+
+public class ConsortiumConfiguration {
```
VRohach, line 3: @RomanChernetskyi  Let's add some documentation for the new classes.

VRohach (approved):

VRohach:
@RomanChernetskyi  Update NEWS please.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/AbstractMarcMatchEventHandler.java
```diff
@@ -80,22 +89,55 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload p
...
+        if (consortiumConfigurationOptional.isPresent() && !consortiumConfigurationOptional.get().getCentralTenantId().equals(payload.getTenant())) {
+          LOG.debug("matchCentralTenantIfNeededAndCombineWithLocalMatchedRecords:: Matching on centralTenant with id: {}",
+            consortiumConfigurationOptional.get().getCentralTenantId());
+          return retrieveMarcRecords(matchField, consortiumConfigurationOptional.get().getCentralTenantId())
+            .onSuccess(result -> {
+              if (!result.isEmpty()) {
+                payload.getContext().put(CENTRAL_TENANT_ID, consortiumConfigurationOptional.get().getCentralTenantId());
+              }
+            })
+            .map(centralTenantResult -> Stream.concat(recordList.stream(), centralTenantResult.stream()).toList());
```
okolawole-ebsco, line 120: Is there a tenant identifier on a record? Thinking it may be good to know which records belong to the central tenant or member tenant.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/MarcAuthorityMatchEventHandler.java
```diff
@@ -16,14 +18,18 @@
...
+      DI_SRS_MARC_AUTHORITY_RECORD_NOT_MATCHED, consortiumConfigurationCache, vertx);
   }
 
   @Override
   public boolean isPostProcessingNeeded() {
     return false;
   }
 
+  @Override
+  boolean isConsortiumAvailable() {
```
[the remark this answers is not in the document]
RomanChernetskyi: We use this method to define if we should initiate a consortium process for a particular type of file, as far as we are using the same abstract class for each file type (MARC BIB, MARC AUTHORITY, MARC HOLDINGS). For now, we should start the consortium process only for MARC BIB, so for MARC AUTHORITY and MARC HOLDINGS 'isConsortiumAvailable' is false

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/AbstractMarcMatchEventHandler.java
RomanChernetskyi: We do not have such an identifier

mod-source-record-storage-server/src/main/java/org/folio/services/entities/ConsortiumConfiguration.java
RomanChernetskyi: Thanks, done

mod-source-record-storage-server/src/main/java/org/folio/services/caches/ConsortiumConfigurationCache.java
RomanChernetskyi: Thanks, done

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/MarcAuthorityMatchEventHandler.java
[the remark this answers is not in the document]
okolawole-ebsco: Thanks! I was thinking we should have a better name, but i couldn't come up with something I like so we will leave as is.

PR #569, opened 2023-09-18
MODINV-824/MODSOURCE-690: Make changes in SRS post processing handler to update MARC for shared Instance

VRohach:
## Purpose
To unable update MARC-record on the central tenant if the Instance was updated on the previous step for Consortium-flow (in the ReplaceInstanceEventHandler in the mod-inventory)

## Approach
Replace local tenantId on the centralTenantId if needed,
Tests added.
 
## Learning
JIRA: https://issues.folio.org/browse/MODSOURCE-690

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
@@ -72,6 +72,8 @@ public abstract class AbstractPostProcessingEventHandler implements EventHandler
   private static final String FAILED_UPDATE_STATE_MSG = "Error during update records state to OLD";
   private static final String ID_FIELD = "id";
   public static final String POST_PROCESSING_INDICATOR = "POST_PROCESSING";
+  public static final String CENTRAL_TENANT_INSTANCE_UPDATED_FLAG = "CENTRAL_TENANT_INSTANCE_UPDATED";
+  public static final String CENTRAL_TENANT_ID = "CENTRAL_TENANT_ID";
   private final KafkaConfig kafkaConfig;
   private final MappingParametersSnapshotCache mappingParamsCache;
   private final Vertx vertx;
@@ -95,7 +97,17 @@ public abstract class AbstractPostProcessingEventHandler implements EventHandler
         .compose(parametersOptional -> parametersOptional
           .map(mappingParams -> prepareRecord(dataImportEventPayload, mappingParams))
           .orElse(Future.failedFuture(format(MAPPING_PARAMS_NOT_FOUND_MSG, jobExecutionId))))
-        .compose(record -> saveRecord(record, dataImportEventPayload.getTenant()))
+        .compose(record -> {
+          if (dataImportEventPayload.getContext().get(CENTRAL_TENANT_INSTANCE_UPDATED_FLAG) != null &&
+            dataImportEventPayload.getContext().get(CENTRAL_TENANT_INSTANCE_UPDATED_FLAG).equals("true")) {
+            String centralTenantId = dataImportEventPayload.getContext().get(CENTRAL_TENANT_ID);
+            dataImportEventPayload.getContext().remove(CENTRAL_TENANT_INSTANCE_UPDATED_FLAG);
+            dataImportEventPayload.getContext().remove(CENTRAL_TENANT_ID);
+            LOG.info("handle:: Processing AbstractPostProcessingEventHandler - saving record by jobExecutionId: {} for the central tenantId: {}",jobExecutionId, centralTenantId);
+            return saveRecord(record, centralTenantId);
+          }
+          return saveRecord(record, dataImportEventPayload.getTenant());
+        })
         .onSuccess(record -> {
           sendReplyEvent(dataImportEventPayload, record);
           sendAdditionalEvent(dataImportEventPayload, record);
@@ -124,6 +136,7 @@ public abstract class AbstractPostProcessingEventHandler implements EventHandler
   }
 
   protected abstract void sendAdditionalEvent(DataImportEventPayload dataImportEventPayload, Record record);
+
   protected abstract String getNextEventType(DataImportEventPayload dataImportEventPayload);
 
   protected String getEventKey() {
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
 * [MODSOURCE-636](https://issues.folio.org/browse/MODSOURCE-636) Implement async migration service
 * [MODSOURCE-674](https://issues.folio.org/browse/MODSOURCE-674) Ensure only one background job can be triggered to clean up outdated marc indexers
 * [MODSOURCE-530](https://issues.folio.org/browse/MODSOURCE-530) Fix duplicate records in incoming file causes problems after overlay process with no error reported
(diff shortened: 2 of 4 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
```diff
@@ -72,14 +74,19 @@ public abstract class AbstractPostProcessingEventHandler implements EventHandler
   private static final String FAILED_UPDATE_STATE_MSG = "Error during update records state to OLD";
   private static final String ID_FIELD = "id";
   public static final String POST_PROCESSING_INDICATOR = "POST_PROCESSING";
+  public static final String CENTRAL_TENANT_INSTANCE_UPDATED_FLAG = "CENTRAL_TENANT_INSTANCE_UPDATED";
```
okolawole-ebsco, line 77: In general, I don't like these special values straddling mod-inventory and SRS. Given how DI works today, it is necessary to propagate the appropriate data in the context between the two modules.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
```diff
@@ -289,13 +302,18 @@ private void executeHridManipulation(Record record, JsonObject externalEntity) {
    * @param tenantId - tenantId
    * @return - Future with Record result
    */
-  private Future<Record> saveRecord(Record record, String tenantId) {
+  private Future<Record> saveRecord(Record record, String tenantId, String centralTenantId) {
```
okolawole-ebsco, line 305: Something seems off with this method. It should only accept a Record and a tenant Id.

There shouldn't be a corresponding source record for the shadow instance in the member tenant so the source record should never be present and no need to update the parsed record. I think you should invoke the copy of the snapshot outside of this method and call it right before calling this method.

mod-source-record-storage-server/src/main/java/org/folio/services/SnapshotService.java
```diff
@@ -62,4 +62,14 @@ Future<SnapshotCollection> getSnapshots(Condition condition, Collection<OrderFie
...
\ No newline at end of file
+  /**
+   * Copy {@link Snapshot} to other tenant
+   *
+   * @param snapshotId       Snapshot id
+   * @param sourceTenantId source tenant id
+   * @param targetTenantId target tenant id
+   * @return future with copied Snapshot
+   */
+  Future<Snapshot> copySnapshotToOtherTenant(String snapshotId, String sourceTenantId, String targetTenantId);
```
okolawole-ebsco, line 73: Please educate me about snapshots, they are really "job executions" from SRM correct?

Updating a central tenant instance means that it already exists against the original snapshot that created it. Updating the central tenant instance means adding a new generation for the record right? if so, do we need to declare a new snapshot?
Thinking the current code will create snapshots in SRS that don't have corresponding job executions in SRM.

RomanChernetskyi (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/SnapshotService.java
VRohach: Yes, you are right. 
So the main thing of this logic - is that we can't manipulate the record (in this way save record) if we don't have a snapshot for it. And in our consortium scope - if we want to save a record on the central tenant - it wont exist there (because there is NO data-import job there, it will be only on the member-tenant where di-job was executed). We discussed with back-end team and decided to just copy the snapshot from the member tenant to the central one, if there is consortium manipulations. So we won't create jobExecution/job info or smth else, this will be just kind of a stub snapshot with stub data to be able to manipulate records in the scope of the central tenant. Feel free to ask any questions.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
VRohach: @okolawole-ebsco  Thank you for your comment!
I agree with you.
Actually, I changed this logic and moved it to the higher level.
Currently, this logic regarding snapshots is inside my method "saveRecordForCentralTenant".
And I left saveRecord() one with the default implementation. 
Feel free to review it!

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
VRohach: @okolawole-ebsco  So we decided to provide these "indicators" in the payload context and they will be removed after using. (See saveRecordForCentralTenant() method). 
I don't understand you correctly. 
Could you please describe - where should we provide these values? If not inside payloadContext? Thank you!

okolawole-ebsco: I dont have any other suggestions. I understand why it was done. When we refactor some stuff in the coming release, we might come up with something better. No actions neccessary.

okolawole-ebsco: Where I am thinking is that we would put something like this in data-import-processing-core.

mod-source-record-storage-server/src/main/java/org/folio/services/SnapshotService.java
okolawole-ebsco: Is a snapshot needed to create another generation of the source record?

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
VRohach: @okolawole-ebsco  Oh, I see. I remember it.

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/AbstractPostProcessingEventHandler.java
okolawole-ebsco: Looks good!

mod-source-record-storage-server/src/main/java/org/folio/services/SnapshotService.java
VRohach: @okolawole-ebsco  As far as I can see here in scope of code there are no operations regarding snapshots while updating record (creating another generation). But it can assumed that it is already in the record.

okolawole-ebsco (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/SnapshotService.java
VRohach: @okolawole-ebsco  It seems like there is the case when it won't be exits(I tested it already): when we create an Instance manually (not via data-import) on the member-tenant. After that, we will share it to the central one. And trying to update it via data-import on the central tenant. So the source of the Instance will be = FOLIO(after manually creation) but after mapping inside di-core it will be MARC (during data-import job). And this flow will look into srs to retrieve MARC-record (because source=MARC not FOLIO), and cant find this MARC-record on the central tenant. That's why there will be created brand new MARC-record in this case.

PR #586, opened 2023-12-19
MODSOURCE-707 Handle authority domain delete events

psmagin:
## Purpose
This story focuses on the consumption of authority domain events within the mod-source-record-storage module. It involves processing these events based on specific conditions. On a soft-delete event, the corresponding record will be marked with the "DELETED" state. On a hard-delete event, the record will be completely deleted from the database. Events of other domain types will be ignored.

## Approach
- create a new verticle for consumption of domain events (verticles for consumers were refactored, common logic moved to absract class)
- create event hadler
- removed scheduled removing of records with DELETE state

## Learning
https://issues.folio.org/browse/MODSOURCE-707


mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
```diff
@@ -0,0 +1,103 @@
...
+  public AuthorityDomainKafkaHandler(RecordService recordService) {
+    this.recordService = recordService;
+  }
+
+  @Override
+  public Future<String> handle(KafkaConsumerRecord<String, String> consumerRecord) {
+    log.trace("handle:: Handling kafka record: '{}'", consumerRecord);
+    String authorityId = consumerRecord.key();
+    if (isUnexpectedDomainEvent(consumerRecord)) {
+      log.trace("handle:: Expected only {} domain type. Skipping authority domain kafka record [ID: '{}']",
```
okolawole-ebsco, line 39: it the domain event is unexpected, should a `return` be here?

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
```diff
@@ -0,0 +1,103 @@
...
+          return Future.succeededFuture();
+        }
+        var matchedId = recordCollection.getRecords().get(0).getMatchedId();
+
+        return recordService.updateRecordsState(matchedId, RecordState.DELETED, RecordType.MARC_AUTHORITY, tenantId);
+      }).map(authorityId);
+  }
+
+  private Future<String> performHardDelete(String authorityId, String tenantId) {
+    return recordService.deleteRecordsByExternalId(authorityId, tenantId).map(authorityId);
```
okolawole-ebsco, line 70: is authority id and external Id the same thing?

mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/AbstractConsumerVerticle.java
```diff
@@ -0,0 +1,92 @@
...
+      subscriptionDefinition = SubscriptionDefinition.builder()
+        .eventType(eventType)
+        .subscriptionPattern(createSubscriptionPattern(kafkaConfig.getEnvId(), eventType))
+        .build();
+    }
+    return subscriptionDefinition;
+  }
+
+  private String getConsumerName() {
+    return constructModuleName() + "_" + getClass().getSimpleName();
```
okolawole-ebsco, line 90: Does this include the module version?

psmagin: Yes

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
psmagin: Yes

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
psmagin: Yes, fixed.

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
okolawole-ebsco: Is it possible for an authority to have an external id that is not the authority id?

okolawole-ebsco (approved):

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
psmagin: I don't think such case exists.

TsaghikKhachatryan (approved):

viacheslavkol (approved):

PR #587, opened 2023-12-28
MODSOURMAN-1106 The status of Instance is '-' in the Import log after uploading file. The numbers of updated SRS and Instance are not displayed in the Summary table.

Maksat-Galymzhan:
## Purpose
MODSOURMAN-1106 The status of Instance is '-' in the Import log after uploading file. The numbers of updated SRS and Instance are not displayed in the Summary table.

## Approach
1. Changed next event type of post processing marc bib during update action.
2. Fixed tests.

## Learning
[MODSOURMAN-1106](https://issues.folio.org/browse/MODSOURMAN-1106)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
@@ -123,10 +123,7 @@ public abstract class AbstractUpdateModifyEventHandler implements EventHandler {
           }
           return recordService.saveRecord(changedRecord, payload.getTenant());
         })
-        .onSuccess(savedRecord -> {
-          payload.setEventType(getNextEventType());
-          future.complete(payload);
-        })
+        .onSuccess(savedRecord -> submitSuccessfulEventType(payload, future))
         .onFailure(throwable -> {
           LOG.warn("handle:: Error while MARC record modifying", throwable);
           future.completeExceptionally(throwable);
@@ -136,6 +133,11 @@ public abstract class AbstractUpdateModifyEventHandler implements EventHandler {
       future.completeExceptionally(e);
     }
     return future;
+  }
+
+  protected void submitSuccessfulEventType(DataImportEventPayload payload, CompletableFuture<DataImportEventPayload> future) {
+    payload.setEventType(getUpdateEventType());
+    future.complete(payload);
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


RomanChernetskyi (approved):

Aliaksandr-Fedasiuk (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractUpdateModifyEventHandler.java
```diff
@@ -124,7 +124,7 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload p
           return recordService.saveRecord(changedRecord, payload.getTenant());
         })
         .onSuccess(savedRecord -> {
-          payload.setEventType(getNextEventType());
+          payload.setEventType(getUpdateEventType());
```
KaterynaSenchenko, line 127: If the success handling is the only thing that is different from the abstract method implementation, maybe it is worth extracting it in a separate method and overriding it for each handler. It should reduce the code duplication

yaroslav-epam (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/MarcBibUpdateModifyEventHandler.java
```diff
@@ -88,6 +96,16 @@ protected EntityType modifiedEntityType() {
...
   }
 
+  @Override
+  protected void submitSuccessfulEventType(DataImportEventPayload payload, CompletableFuture<DataImportEventPayload> future, MappingDetail.MarcMappingOption marcMappingOption) {
+    super.submitSuccessfulEventType(payload, future, marcMappingOption);
+
+    if (marcMappingOption.value().equals(MODIFY.value())) {
+      payload.setEventType(getModifyEventType());
+    }
+    future.complete(payload);
```
KaterynaSenchenko, line 106: the future is completed already at this point, isn't it?

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/MarcBibUpdateModifyEventHandler.java
```diff
@@ -88,6 +98,17 @@ protected EntityType modifiedEntityType() {
     return MARC_BIBLIOGRAPHIC;
   }
 
+  @Override
+  protected void submitSuccessfulEventType(DataImportEventPayload payload, CompletableFuture<DataImportEventPayload> future, MappingDetail.MarcMappingOption marcMappingOption) {
+    if (marcMappingOption.value().equals(MODIFY.value())) {
+      payload.setEventType(getModifyEventType());
+    }
+    if (marcMappingOption.value().equals(UPDATE.value())) {
```
yaroslav-epam, line 106: looks like conditions could be united via `||`

KaterynaSenchenko (approved):

PR #588, opened 2024-01-04
[MODSOURCE-731] Add PUT endpoint to update SRS record

RomanChernetskyi:
## Purpose
Add PUT endpoint to update SRS record

## Approach
Added endpoint with URL `/source-storage/records/{id}/generation` Basically new endpoint execute 'saveRecord' method but only for record with a matched id that already exists at DB, so we are creating a new ACTUAL record with incremented generation

## Learning
https://issues.folio.org/browse/MODSOURCE-731

### DIFF
--- a/mod-source-record-storage-server/src/test/java/org/folio/rest/impl/RecordApiTest.java
+++ b/mod-source-record-storage-server/src/test/java/org/folio/rest/impl/RecordApiTest.java
@@ -658,6 +658,33 @@ public class RecordApiTest extends AbstractRestVerticleTest {
   }
 
   @Test
+  public void shouldSendBadRequestWhenUpdateRecordGenerationWithDuplicate(TestContext testContext) {
+    postSnapshots(testContext, snapshot_1);
+
+    Async async = testContext.async();
+    Response createResponse = RestAssured.given()
+      .spec(spec)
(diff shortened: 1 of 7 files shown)


Aliaksandr-Fedasiuk (approved):

Maksat-Galymzhan (approved):

mod-source-record-storage-server/src/test/java/org/folio/rest/impl/RecordApiTest.java
```diff
@@ -589,6 +598,121 @@ public void shouldUpdateErrorRecordOnPut(TestContext testContext) {
...
+    RestAssured.given()
+      .spec(spec)
+      .body(recordForUpdate)
+      .when()
+      .put(SOURCE_STORAGE_RECORDS_PATH + "/" + createdRecord.getMatchedId() + "/" + GENERATION)
+      .then()
+      .statusCode(HttpStatus.SC_BAD_REQUEST);
+    async.complete();
+  }
+
```
yaroslav-epam, line 685: extra line

yaroslav-epam (approved):

KaterynaSenchenko (approved):

PR #590, opened 2024-01-08
MODSOURCE-717: MARC modifications not processed when placed after Holdings Update action in a job profile

Maksat-Galymzhan:
## Purpose
MARC modifications not processed when placed after Holdings Update action in a job profile.

## Approach
ADDED holding update subscription. Also it is checked behavior with item case.

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/DataImportConsumersVerticle.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/DataImportConsumersVerticle.java
@@ -5,6 +5,7 @@ import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_INVENTORY_AUTHO
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_INVENTORY_HOLDINGS_CREATED_READY_FOR_POST_PROCESSING;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_INVENTORY_HOLDING_CREATED;
 import static org.folio.rest.jaxrs.model.DataImportEventTypes.DI_INVENTORY_HOLDING_MATCHED;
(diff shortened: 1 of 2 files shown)


Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

JavokhirAbdullayev (approved):

mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/DataImportConsumersVerticle.java
```diff
@@ -42,6 +43,7 @@ public class DataImportConsumersVerticle extends AbstractConsumerVerticle {
     DI_INVENTORY_AUTHORITY_CREATED_READY_FOR_POST_PROCESSING.value(),
     DI_INVENTORY_AUTHORITY_UPDATED_READY_FOR_POST_PROCESSING.value(),
     DI_INVENTORY_HOLDING_CREATED.value(),
+    DI_INVENTORY_HOLDING_UPDATED.value(),
```
KaterynaSenchenko, line 46: What about  DI_INVENTORY_ITEM_UPDATED? Do we need to add this subscription as well? For the case when Modify action is placed after the Update Item action

Maksat-Galymzhan: It's already been added.  [   DI_INVENTORY_ITEM_UPDATED.value(),](https://github.com/folio-org/mod-source-record-storage/blob/77eda686feabf143d9922e170f14cea9895f1b29/mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/DataImportConsumersVerticle.java#L56)

KaterynaSenchenko (approved):

PR #591, opened 2024-01-09
MODSOURCE-733 Reduce memory allocations of Strings

okolawole-ebsco:
## Purpose
Reduce memory allocations of strings to reduce memory requirements of SRS and improve performance at the limit

## Approach
Prior to this change, byte arrays obtained from the Kafka clients are converted to Strings before being passed to a consumer handler. The String is then converted to a POJO for further logic. Due to requirements of data import, kafka payloads can be large. For example, if payload for a parsed MARC record is 1MB(this is a common occurrence), its string representation will be 1MB + other bytes.

After this change, byte arrays are coverted directly to POJO so intermediate Strings are not created. Internally, Jackson streams the byte array rather than create a String representation. This means that memory used is not directly proportional to the size of the kafka payloads encountered.

This is a problem in SRS because large strings are created. In production, large payloads containing parsed marc records and current node of complex Job profiles exists.

### DIFF
--- a/mod-source-record-storage-server/src/main/resources/vertx-default-jul-logging.properties
+++ b/mod-source-record-storage-server/src/main/resources/vertx-default-jul-logging.properties
@@ -1,5 +1,0 @@
-handlers = java.util.logging.ConsoleHandler
-.level = ALL
-
-logger.cql2pgjson.level = ERROR
-logger.cql2pgjson.name = org.folio.cql2pgjson.CQL2PgJSON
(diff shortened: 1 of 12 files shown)


mod-source-record-storage-server/src/main/resources/vertx-default-jul-logging.properties
okolawole-ebsco, line 1: This file was removed because it is causing JOOQ's reactive query executor to log all sql statements but those logs are not visible because log4j2 configuration already exists. So the cost is incurred, but no value since they don't show up. Removing this configuration means less strings are being allocated for log statements.

psmagin (approved):

KaterynaSenchenko (approved):

RuslanLavrov (approved):

yaroslav-epam (approved):

PR #594, opened 2024-02-01
MODSOURCE-729 - Implement new endpoint to be used for matching

RuslanLavrov:
## Purpose
to implement new endpoint to be used for MARC records matching

## Approach
* implement `POST /source-storage/records/matching` endpoint that consumes matching conditions from request body and provides pagination capabilities. The endpoint returns RecordsIdentifiersCollection containing list of pairs (recordId, externalId) in a response.
* add tests

## Learning
[MODSOURCE-729](https://issues.folio.org/browse/MODSOURCE-729)

Requires: PR https://github.com/folio-org/data-import-raml-storage/pull/306


KaterynaSenchenko (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -491,6 +494,68 @@ private void appendJoinAlternative(SelectJoinStep<?> selectJoinStep, ParseLeader
...
 
+  @Override
+  public Future<RecordsIdentifiersCollection> getMatchedRecordsIdentifiers(MatchField matchedField, TypeConnection typeConnection,
+                                                                           boolean externalIdRequired, int offset, int limit, String tenantId) {
+    Table<org.jooq.Record> marcIndexersPartitionTable = table(name(MARC_INDEXERS_PARTITION_PREFIX + matchedField.getTag()));
+    if (matchedField.getValue() instanceof MissingValue) {
+      return Future.succeededFuture(new RecordsIdentifiersCollection().withTotalRecords(0));
+    }
+
+    return getQueryExecutor(tenantId).transaction(txQE -> txQE.query(dsl -> {
```
okolawole-ebsco, line 505: Are there facilities to disable returning a count as part of a API response. If so, we should make the count call to be optional.

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -491,6 +494,68 @@ private void appendJoinAlternative(SelectJoinStep<?> selectJoinStep, ParseLeader
     }
   }
 
+  @Override
+  public Future<RecordsIdentifiersCollection> getMatchedRecordsIdentifiers(MatchField matchedField, TypeConnection typeConnection,
+                                                                           boolean externalIdRequired, int offset, int limit, String tenantId) {
+    Table<org.jooq.Record> marcIndexersPartitionTable = table(name(MARC_INDEXERS_PARTITION_PREFIX + matchedField.getTag()));
```
okolawole-ebsco, line 500: Is it possible that more than one tag would need to be matched?

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
okolawole-ebsco: Also, when mod-inventory makes a call to do matching, it should exclude the count.

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -300,11 +313,63 @@ public Future<Void> updateRecordsState(String matchedId, RecordState state, Reco
...
+        .collect(collectingAndThen(toList(), identifiers -> new RecordsIdentifiersCollection()
+          .withIdentifiers(identifiers).withTotalRecords(recordCollection.getTotalRecords()))));
+  }
+
   private Future<Record> setMatchedIdForRecord(Record record, String tenantId) {
     String marcField999s = getFieldFromMarcRecord(record, TAG_999, INDICATOR, INDICATOR, SUBFIELD_S);
     if (marcField999s != null) {
       // Set matched id from 999$s marc field
-      LOG.debug(format("setMatchedIdForRecord:: Set matchedId: %s from 999$s field for record with id: %s", marcField999s, record.getId()));
+      LOG.debug("setMatchedIdForRecord:: Set matchedId: {} from 999$s field for record with id: {}", marcField999s, record.getId());
```
okolawole-ebsco, line 372: Good change!

mod-source-record-storage-server/src/test/java/org/folio/rest/impl/RecordsMatchingApiTest.java
```diff
@@ -0,0 +1,375 @@
...
+public class RecordsMatchingApiTest extends AbstractRestVerticleTest {
+
+  private static final String RECORDS_MATCHING_PATH = "/source-storage/records/matching";
+  private static final String MARC_BIB_PARSED_CONTENT_WITH_ADDITIONAL_FIELDS = "{\"leader\":\"01589ccm a2200373   4500\",\"fields\":[{ \"001\": \"12345\" }, {\"007\": \"12569\"},{\"007\": \"1234567\"},{\"024\": {\"ind1\": \"8\", \"ind2\": \"0\", \"subfields\": [{\"a\": \"test123\"}]}}, {\"024\": {\"ind1\": \"1\", \"ind2\": \"1\", \"subfields\": [{\"a\": \"test45\"}]}},{\"035\": {\"ind1\": \" \", \"ind2\": \" \", \"subfields\": [{\"a\": \"nin00009530412\"}]}}, {\"035\": {\"ind1\": \" \", \"ind2\": \" \", \"subfields\": [{\"a\": \"12345\"}]}},{\"245\":{\"ind1\":\"1\",\"ind2\":\"0\",\"subfields\":[{\"a\":\"Neue Ausgabe sämtlicher Werke,\"}]}},{\"948\":{\"ind1\":\"\",\"ind2\":\"\",\"subfields\":[{\"a\":\"acf4f6e2-115c-4509-9d4c-536c758ef917\"},{\"b\":\"681394b4-10d8-4cb1-a618-0f9bd6152119\"},{\"d\":\"12345\"},{\"e\":\"lts\"},{\"x\":\"addfast\"}]}},{\"999\":{\"ind1\":\"f\",\"ind2\":\"f\",\"subfields\":[{\"s\":\"acf4f6e2-115c-4509-9d4c-536c758ef917\"}, {\"i\":\"681394b4-10d8-4cb1-a618-0f9bd6152119\"}]}}]}";
```
okolawole-ebsco, line 42: nitpick: consider having this JSON objects as a resource that is loaded into tests. It will make the JSON easier to read if it is prettified and it keeps the large JSON object separate from test code.

okolawole-ebsco:
Nice to see good javadocs on your methods!

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
RuslanLavrov: Currently, such a scenario is not expected since a match profile supports only one matching criteria.

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
RuslanLavrov: It is possible not to return totalRecords field in a response by setting it to null. Assume that if it's necessary, the logic to count and return "totalRecords" field can be managed by specifying "returnTotalRecords" query parameter in a request that would be true by default.


> Also, when mod-inventory makes a call to do matching, it should exclude the count.

Why mod-inventory should exclude the total records count? How would mod-inventory know that it has received identifiers for all records that match the matching criteria and doesn't need to fetch next batch? As far as understand mod-inventory should get identifiers of all matched records when such records' identifiers are needed for subsequent processing of instances matching.

VRohach (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -300,11 +313,63 @@ public Future<Void> updateRecordsState(String matchedId, RecordState state, Reco
...
+      return processDefaultMatchField(matchField, tenantId, typeConnection,
+        recordMatchingDto.getOffset(), recordMatchingDto.getLimit());
+    }
+    return recordDao.getMatchedRecordsIdentifiers(matchField, typeConnection, true,
+      recordMatchingDto.getOffset(), recordMatchingDto.getLimit(), tenantId);
+  }
+
+  private MatchField prepareMatchField(RecordMatchingDto recordMatchingDto) {
+    // only one matching filter is expected in the current implementation for processing records matching
+    Filter filter = recordMatchingDto.getFilters().get(0);
```
okolawole-ebsco, line 331: We should throw an exception here, if the number of filters is greater than 1.

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
RuslanLavrov: > Will the current state of the code handle this?

The current state will not handle it. I just described a possible approach how it can be implemented. Are you ok with the described approach for making total count calculation optional?

> That is a good point about next batch. The main reason to have the total count configurable is for performance. I assume that matching via kafka messages didn't need to make a call to get total count of the database. I think this shift to REST api to do matching could cause performance degradation due to extra calls to support the CQL/FOLIO HTTP response standard. There should be some way to know that there are more batch, that may not play with FOLIO for now.

Yes, the previous marc-marc matching approach did not need to get the total count because multiple matched records were treated as error. During marc-marc matching feature discussion we talked that multiple results from marc matching will be used in case of subsequent processing of instances matching. I'm thinking maybe mod-inventory could get total records count only by first request to records matching API, and for subsequent requests mod-inventory could specify query parameter not to perform total records calculation.

RuslanLavrov: changed the logic for total records calculation to optional but by introducing "returnTotalRecordsCount" field in the request body instead of "returnTotalRecords" query parameter as I described above.

PR #595, opened 2024-02-02
MODSOURCE-739 Create Kafka topics instead of relying on auto create in mod-srs

JavokhirAbdullayev:
## Purpose
Create Kafka topics instead of relying on auto create in mod-srs [MODSOURCE-739](https://issues.folio.org/browse/MODSOURCE-739)

## Approach
* create kafka topics during tenant init
* add tests

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/SRSKafkaTopicService.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/SRSKafkaTopicService.java
@@ -1,5 +1,8 @@
 package org.folio.services;
 
+import static org.folio.kafka.KafkaTopicNameHelper.formatTopicName;
+import static org.folio.kafka.services.KafkaEnvironmentProperties.environment;
+
 import org.folio.kafka.services.KafkaTopic;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.PropertySource;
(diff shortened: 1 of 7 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/SRSKafkaTopicService.java
```diff
@@ -0,0 +1,103 @@
...
+  private Integer diMarcAuthorityRecordMatchedPartitions;
+
+  @Value("${di_marc_authority_record_not_matched.partitions}")
+  private Integer diMarcAuthorityRecordNotMatchedPartitions;
+
+  @Value("${di_marc_authority_record_deleted.partitions}")
+  private Integer diMarcAuthorityRecordDeletedPartitions;
+
+  public KafkaTopic[] createTopicObjects() {
+    var marcBib = new SRSKafkaTopic("MARC_BIB", marcBibPartitions);
```
dmytrokrutii, line 46: Perhaps it would be better to simplify to
```
return new SRSKafkaTopic[]{
          new SRSKafkaTopic("MARC_BIB", marcBibPartitions),
          new SRSKafkaTopic("DI_PARSED_RECORDS_CHUNK_SAVED", diParsedRecordsChunkSavedPartitions),
          new SRSKafkaTopic("DI_SRS_MARC_BIB_INSTANCE_HRID_SET", diSrsMarcBibInstanceHridSetPartitions),
          new SRSKafkaTopic("DI_SRS_MARC_BIB_RECORD_MODIFIED", diSrsMarcBibRecordModifiedPartitions),
          ....
  };
  ```

dmytrokrutii (approved):

VRohach (approved):

PR #609, opened 2024-03-21
MODSOURCE-757 Expose DB Reconnection Attempt Config Options

okolawole-ebsco:
## Purpose
Counter io.netty.channel.StacklessClosedChannelException exceptions that are generated intermittently.

## Approach
Allow reconnect attempts configuration to be configured for the Vert.x postgres client. Also, include a default reconnect attempt of 3 to the module descriptor.

## Learning
https://folio-org.atlassian.net/browse/MODSOURCE-757


mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
okolawole-ebsco, line 1: Ideally, configuration keys defined in the class ought to reference the keys defined in RMB rather than duplicating them here. If the keys were to change in RMB, there would be a disconnect here.

RomanChernetskyi (approved):

KaterynaSenchenko (approved):

PR #614, opened 2024-03-29
MODSOURCE-742 Configure Postgres SSL connection

dmytrokrutii:
## Purpose
PostgreSQL SSL connection does not work

## Approach
Configure PostgreSQL SSL connection

## Learning
[MODSOURCE-742](https://folio-org.atlassian.net/browse/MODSOURCE-742)


Maksat-Galymzhan (approved):

KaterynaSenchenko (approved):

RomanChernetskyi (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java
```diff
@@ -221,16 +218,7 @@ private static PgPool getCachedPool(Vertx vertx, String tenantId) {
...
-      .setHost(postgresConfig.getString(HOST))
-      .setPort(postgresConfig.getInteger(PORT))
-      .setDatabase(postgresConfig.getString(DATABASE))
-      .setUser(postgresConfig.getString(USERNAME))
-      .setPassword(postgresConfig.getString(PASSWORD))
-      .setIdleTimeout(postgresConfig.getInteger(IDLE_TIMEOUT, 60000))
-      .setIdleTimeoutUnit(TimeUnit.MILLISECONDS)
-      .setReconnectAttempts(postgresConfig.getInteger(DB_RECONNECTATTEMPTS, 0))
-      .setReconnectInterval(postgresConfig.getLong(DB_RECONNECTINTERVAL, 1L))
+    return PgConnectOptionsHelper.createPgConnectOptions(postgresConfig)
```
RuslanLavrov, line 221: looks like we also need to provide SSL configuration for dataSource pool in case SERVER_PEM is set
https://github.com/folio-org/mod-source-record-storage/blob/267a0ff5d2a55daca86ec43ff2910d00918d90a9/mod-source-record-storage-server/src/main/java/org/folio/dao/PostgresClientFactory.java#L237

PR #624, opened 2024-05-21
MODSOURCE-773: MARC Search omits suppressed from discovery records in default search.

VRohach:
## Purpose
Create an opportunity to get the response that will include all records regardless of their discovery suppress status.
## Approach
Default value "false" for target field removed.
boolean change to Boolean to be able to accept null values (the same behavior as for limit/offset)

## Learning
JIRA: https://folio-org.atlassian.net/browse/MODSOURCE-773


VRohach:
This PR should be merged ONLY after https://github.com/folio-org/data-import-raml-storage/pull/310 will be merged.

Maksat-Galymzhan (approved):

JavokhirAbdullayev (approved):

RuslanLavrov (approved):

PR #626, opened 2024-06-04
MODSOURCE-773: Remove default searching via "deleted"="false".

VRohach:
## Purpose
The main goal is to remove default searching if "deleted" parameter is NOT set inside the request.It will find either deleted=true or deleted=false.

## Approach
Schema changed+additional condition added - if isDeleted==null, then do not add a condition to the CQL query.
Test added.

## Learning
https://folio-org.atlassian.net/browse/MODSOURCE-773


VRohach:
This PR should be merged ONLY after https://github.com/folio-org/data-import-raml-storage/pull/311 will be merged!

JavokhirAbdullayev (approved):

KaterynaSenchenko (approved):

Maksat-Galymzhan (approved):

RuslanLavrov (approved):

dmytrokrutii (approved):

PR #628, opened 2024-06-20
MODSOURCE-763 SRS MARC query search won't allow you to create two search conditions for a single field (SPIKE)

Maksat-Galymzhan:
## Introduced changes
- changed the approach to refer to marc_indexers table. Previously, there have been used partitions of the marc_indexers table, and not to boiler-plate script with joins, there is introduced field number conditioning inside of search criteria.
- Due to complex structure of the criteria search, there are numerous cases which need to be considered, such as parenthesis included search. E.g., (050.a ^= 'a' and 050.b ^= 'b') or 050.c ^= 'c', or search criteria relates date search.

## Actions to do
- Please notice TODO's in the java code.
- Main problem is to parse conditions with additional parenthesis to not break condition wholeness. There is a test which is ignored, due to the mentioned issue.
- Not added **marc_indexers** tracking, during the investigation no noticeable changes found after removing this join in the script.

## Learning
[MODSOURCE-763](https://folio-org.atlassian.net/browse/MODSOURCE-763)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
@@ -11,6 +11,13 @@ import io.vertx.core.Vertx;
 import io.vertx.reactivex.pgclient.PgPool;
 import io.vertx.reactivex.sqlclient.SqlConnection;
 import io.vertx.sqlclient.Row;
+import net.sf.jsqlparser.JSQLParserException;
+import net.sf.jsqlparser.expression.BinaryExpression;
+import net.sf.jsqlparser.expression.Expression;
+import net.sf.jsqlparser.expression.Parenthesis;
+import net.sf.jsqlparser.expression.operators.conditional.AndExpression;
+import net.sf.jsqlparser.expression.operators.conditional.OrExpression;
+import net.sf.jsqlparser.parser.CCJSqlParserUtil;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.text.StrSubstitutor;
 import org.apache.commons.lang3.StringUtils;
@@ -90,7 +97,6 @@ import java.sql.SQLException;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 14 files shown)


KaterynaSenchenko (approved):

JavokhirAbdullayev (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -420,39 +520,33 @@ public Flowable<Row> streamMarcRecordIds(ParseLeaderResult parseLeaderResult, Pa
...
     /* Join both in one query */
-    String sql = DSL.select().from(searchQuery).rightJoin(countQuery).on(DSL.trueCondition()).getSQL(ParamType.INLINED);
-
+    String sql = "";
+    if (parseFieldsResult.isEnabled()) {
+      sql = DSL.with(commonTableExpression).select().from(searchQuery).rightJoin(countQuery).on(DSL.trueCondition()).getSQL(ParamType.INLINED);
+    } else {
+      sql = select().from(searchQuery).rightJoin(countQuery).on(DSL.trueCondition()).getSQL(ParamType.INLINED);
+    }
+    String finalSql = sql;
```
VRohach, line 532: @Aliaksandr-Fedasiuk Maybe it would be better to log finalSql? In debug level, for example.

VRohach (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
Aliaksandr-Fedasiuk: thanks for the idea!

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -406,11 +419,95 @@ public Flowable<Record> streamRecords(Condition condition, RecordType recordType
...
+    if (parseFieldsResult.isEnabled()) {
+      String cteWhereExpression = buildCteWhereCondition(parseFieldsResult.getWhereExpression());
+
+      Expression expr = CCJSqlParserUtil.parseCondExpression(parseFieldsResult.getWhereExpression());
+      String cteHavingExpression = buildCteDistinctCountCondition(expr);
+
+      commonTableExpression = DSL.name(CTE).as(
+        select(
+          field(MARC_ID))
+          .from(MARC_INDEXERS).join(RECORDS_LB.getName())
```
RuslanLavrov, line 501: propose to return back the join on the marc_records_tracking table when parseFieldsResult exists, to make a search on marc_indexers records with the latest version, as it was before

Aliaksandr-Fedasiuk: completed. thanks!

RuslanLavrov: thanks!

PR #631, opened 2024-07-10
MODSOURMAN-1200 Find record by match id instead record id

dmytrokrutii:
## Purpose
The 'No action' status is displayed in the 'SRS Marc' and 'Instance' columns after importing file for updating record on member tenant

## Approach
Find record by match id instead record id

#### TODOS and Open Questions
- [x] Check logging.
## Learning
[MODSOURMAN-1200](https://folio-org.atlassian.net/browse/MODSOURMAN-1200)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
@@ -182,7 +182,7 @@ public class RecordServiceImpl implements RecordService {
     }
     record.setId(UUID.randomUUID().toString());
 
-    return recordDao.getRecordById(matchedId, tenantId)
+    return recordDao.getRecordByMatchedId(matchedId, tenantId)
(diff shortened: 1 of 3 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -182,7 +182,7 @@ public Future<Record> updateRecordGeneration(String matchedId, Record record, St
     }
     record.setId(UUID.randomUUID().toString());
 
-    return recordDao.getRecordById(matchedId, tenantId)
+    return recordDao.getRecordByMatchedId(matchedId, tenantId)
```
KaterynaSenchenko, line 185: is it possible to cover this with some kind of test?

JavokhirAbdullayev (approved):

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #633, opened 2024-07-18
Extend MARC-MARC search query to account for qualifiers

dmytrokrutii:
## Purpose
Enhance the implementation of the source-storage/records/matching endpoint to account for qualifiers when building a search query to get the record for matching.

#### TODOS and Open Questions
- [x] Check logging.
## Learning
[MODSOURCE-783](https://folio-org.atlassian.net/browse/MODSOURCE-783)


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -352,18 +354,26 @@ public Future<List<Record>> getMatchedRecordsWithoutIndexersVersionUsage(MatchFi
...
     Map<String, String> params = new HashMap<>();
+    var qualifierSearch = false;
     params.put("partition", partition);
     params.put("value", getValueInSqlFormat(matchedField.getValue()));
+    if (matchedField.getQualifierMatch() != null) {
+      qualifierSearch = true;
+      params.put("qualifier", getSqlQualifier(matchedField.getQualifierMatch()));
+    }
+    String sql;
+    Condition condition;
```
JavokhirAbdullayev, line 365: if condition is not used, let's remove it

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
JavokhirAbdullayev, line 1: could you please also replace this line ```import static org.jooq.impl.DSL.*;``` with exact importing classes?

KaterynaSenchenko (approved):

RomanChernetskyi (approved):

RuslanLavrov (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
dmytrokrutii: These conditions are used.
`(matchedField.getQualifierMatch() != null` is used to keep the logic that was before we added the search by qualifiers unchanged
`if (matchedField.isControlField())` this condition was there before, because if it is a control field, the sql query should be formed differently

JavokhirAbdullayev (approved):

PR #635, opened 2024-07-25
MODSOURCE-752: Emit Domain Events For Source Records

PBobylev:
## Purpose
[MODSOURCE-752](https://folio-org.atlassian.net/browse/MODSOURCE-752) Emit Domain Events For Source Records.

## Approach
RecordDomainEventPublisher introduced to be triggered after each record(s) create/update operation in RecordDao.

#### TODOS and Open Questions
- [x] Discuss the approach with code owners
- [x] Add RAML definition for event DTO and use it
- [x] Handle all create/update methods in RecordDao
- [x] Implement configuration param for the module to enable/disable the publishing
- [x] Implement new topic creation
- [x] Fix/adapt existed tests
- [x] Cover new changes with tests


PBobylev:
Discussed with @KaterynaSenchenko, approach is approved and to be implemented.

ramls/source-record-domain-event.json
```diff
@@ -0,0 +1,97 @@
...
+      "required": [
+        "eventTTL",
+        "publishedBy"
+      ]
+    },
+    "eventPayload": {
+      "type": "string",
+      "description": "The source record JSON string"
+    },
+    "tenant": {
```
pkjacob, line 78: Do we need "tenant" and "ts" in the event? These fields are not mentioned in the design doc that Olamide created. https://folio-org.atlassian.net/wiki/spaces/DD/pages/62685275/Source+Record+Domain+Eventing

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -1286,9 +1307,10 @@ private MarcBibCollection toMarcBibCollection(QueryResult result) {
   }
 
   @Override
-  public Future<Record> saveUpdatedRecord(ReactiveClassicGenericQueryExecutor txQE, Record newRecord, Record oldRecord) {
+  public Future<Record> saveUpdatedRecord(ReactiveClassicGenericQueryExecutor txQE, Record newRecord, Record oldRecord, Map<String, String> okapiHeaders) {
     LOG.trace("saveUpdatedRecord:: Saving updated record {}", newRecord.getId());
-    return insertOrUpdateRecord(txQE, oldRecord).compose(r -> insertOrUpdateRecord(txQE, newRecord));
+    return insertOrUpdateRecord(txQE, oldRecord).compose(r -> insertOrUpdateRecord(txQE, newRecord))
```
pkjacob, line 1312: Here we are invoking "insertOrUpdateRecord" for both oldRecord and newRecord. So, I think we should emit the event for both oldRecord and newRecord.

mod-source-record-storage-server/src/main/java/org/folio/consumers/QuickMarcKafkaHandler.java
```diff
@@ -67,22 +65,23 @@ public Future<String> handle(KafkaConsumerRecord<String, String> consumerRecord)
     log.trace("handle:: Handling kafka consumerRecord {}", consumerRecord);
 
     var kafkaHeaders = consumerRecord.headers();
-    var params = new OkapiConnectionParams(kafkaHeadersToMap(kafkaHeaders), vertx);
+    var okapiHeaders = toOkapiHeaders(kafkaHeaders, null);
```
AndreiBordak, line 68: i'd create overloaded version of toOkapiHeaders() for this case:
```
public static Map<String, String> toOkapiHeaders(List<KafkaHeader> kafkaHeaders) {
    return toOkapiHeaders(kafkaHeaders, null);
}
```

ramls/source-record-domain-event.json
PBobylev: These 2 fields are part of Kafka message technical design in the Folio library [folio-service-tools-spring-dev](https://github.com/folio-org/folio-service-tools/blob/master/folio-service-tools-spring-dev/src/main/java/org/folio/spring/tools/kafka/BaseKafkaMessage.java) which we use in mod-linked-data.
I wanted to follow the same here

mod-source-record-storage-server/src/main/java/org/folio/consumers/QuickMarcKafkaHandler.java
PBobylev: agree, done

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
PBobylev: I don't think we should do this: only SRS stores multiple versions of the same entity, right?
Other systems should get update message only, otherwise there will be just 2 update messages with the same recordId one after another

AndreiBordak (approved):

pkjacob (approved):

ramls/source-record-domain-event.json
PBobylev: The file isn't needed anymore, re-used existed Event model, since it's compatible

KaterynaSenchenko (approved):

PR #639, opened 2024-08-09
 MODSOURCE-787: Extend MARC-MARC search query to account for comparison part

Aliaksandr-Fedasiuk:
## Purpose
Enhance implementation of /source-storage/records/matching endpoint to account for comparison part when building a search query to get the record for matching.

## Approach
SQL query generation has been changed to take into account the new conditions

## Learning
[MODSOURCE-787](https://folio-org.atlassian.net/browse/MODSOURCE-787)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
@@ -413,7 +413,8 @@ public class RecordDaoImpl implements RecordDao {
     }
 
     return switch (comparisonPartType) {
-      case ALPHANUMERICS_ONLY -> "regexp_replace(\"{partition}\".\"value\", '[^[:alnum:]]', '', 'g')";
+      //case ALPHANUMERICS_ONLY -> "regexp_replace(\"{partition}\".\"value\", '[^[:alnum:]]', '', 'g')";
+      case ALPHANUMERICS_ONLY -> "regexp_replace(\"{partition}\".\"value\", '[^\\w]|_', '', 'g')";
       case NUMERICS_ONLY -> "regexp_replace(\"{partition}\".\"value\", '[^[:digit:]]', '', 'g')";
       default -> DEFAULT_VALUE;
     };
--- a/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
@@ -411,7 +411,7 @@ public class RecordDaoImpl implements RecordDao {
       return "\"{partition}\".\"value\"";
     } else {
       if (comparisonPartType == ALPHANUMERICS_ONLY) {
-        return "regexp_replace(\"{partition}\".\"value\", '[^a-zA-Z0-9]', '', 'g')";
+        return "regexp_replace(\"{partition}\".\"value\", '[^[:alnum:]]', '', 'g')";
       } else {
         return "regexp_replace(\"{partition}\".\"value\", '[^0-9]', '', 'g')";
       }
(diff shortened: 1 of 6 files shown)


KaterynaSenchenko (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -390,6 +406,18 @@ private Condition getMatchedFieldCondition(MatchField matchedField, String parti
     return condition(sql);
   }
 
+  private static String getComparisonValue(Filter.ComparisonPartType comparisonPartType) {
+    if (comparisonPartType == null) {
+      return "\"{partition}\".\"value\"";
+    } else {
+      if (comparisonPartType == ALPHANUMERICS_ONLY) {
+        return "regexp_replace(\"{partition}\".\"value\", '[^a-zA-Z0-9]', '', 'g')";
```
RuslanLavrov, line 414:
```suggestion
        return "regexp_replace(\"{partition}\".\"value\", '[^[:alnum:]]', '', 'g')";
```
propose to apply the following to avoid replacement/lost of characters that are not included in the a-zA-Z range (e.g. cyryllic characters) in case field of existing record contains such

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/AbstractMarcMatchEventHandler.java
```diff
@@ -137,8 +138,9 @@ private Future<List<Record>> retrieveMarcRecords(MatchField matchField, String t
       LOG.debug("retrieveMarcRecords:: Process default field matching, matchField {}, tenant {}", matchField, tenant);
       return processDefaultMatchField(matchField, tenant).map(RecordCollection::getRecords);
     }
+    //TODO is Filter.ComparisonPartType needed?
```
VRohach, line 141: @Aliaksandr-Fedasiuk  Do we need this TODO comment ?

RuslanLavrov, line 141: looks like it is needed to support ComparisonPartType option for marc-authority to marc-authority records matching

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -390,6 +406,18 @@ private Condition getMatchedFieldCondition(MatchField matchedField, String parti
     return condition(sql);
   }
 
+  private static String getComparisonValue(Filter.ComparisonPartType comparisonPartType) {
```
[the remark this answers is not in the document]
Aliaksandr-Fedasiuk: Thanks @VRohach , I agree with using the switch statement, but I don't agree with separating the lines because, in my opinion, it doesn't make the code more understandable.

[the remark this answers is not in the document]
VRohach: @Aliaksandr-Fedasiuk  Got it. Thanks. Then left the best approach from your side.

VRohach (approved):

JavokhirAbdullayev (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/AbstractMarcMatchEventHandler.java
Aliaksandr-Fedasiuk: completed, thanks

Aliaksandr-Fedasiuk: completed, thanks

RuslanLavrov (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/AbstractMarcMatchEventHandler.java
RuslanLavrov: thanks

PR #643, opened 2024-09-12
MODINV-1002 Disallow update of records originated in Linked Data

KaterynaSenchenko:
## Purpose
Disallow updating records originated in Linked Data 

## Approach
Check for linked data id in 999 ff l subfield and throw an error if it is present

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/exceptions/RecordUpdateException.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/exceptions/RecordUpdateException.java
@@ -1,8 +1,11 @@
 package org.folio.services.exceptions;
 
+/**
+ * Exception to indicate a record update error
+ */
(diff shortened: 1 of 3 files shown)


Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/exceptions/RecordUpdateException.java
```diff
@@ -0,0 +1,8 @@
+package org.folio.services.exceptions;
+
+public class RecordUpdateException extends RuntimeException {
```
VRohach, line 3: @KaterynaSenchenko  Let's add documentation for newly created Exceptions.

VRohach (approved):

PR #649, opened 2024-11-05
MODSOURCE-817: Fix data consistency in handling and updating Marc Bib records for links.instance-authority event

mukhiddin-yusuf:
## Purpose
_Make the handler of instance-authority links events consistent so that no data is lost._

## Approach
- make the handling of events for instance-authority links so that no Marc Bib modifications are lost due to concurrent nature of handling events
- make sure the event of the first in a sequence of updates sent by the producer and then received by the consumer is handled before with persisting the changes of related bibs than the events produced/consumed later

#### TODOS and Open Questions
- [x] Check logging.

Removed the environment variable **_srs.kafka.AuthorityLinkChunkConsumer.loadLimit_** as there are not much sense to have the additional concurrency feature, since we want to strictly follow the order of sent events, retrieve and save the chunk of Marc Bib records and only after that process the next chunk of records from the following event. 

## Learning
Closes: [MODSOURCE-817](https://folio-org.atlassian.net/browse/MODSOURCE-817)


Aliaksandr-Fedasiuk:
@mukhiddin-yusuf , please use capital letters in the branch names.

mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/AuthorityLinkChunkConsumersVerticle.java
```diff
@@ -19,8 +18,7 @@ public class AuthorityLinkChunkConsumersVerticle extends AbstractConsumerVerticl
 
   private final AuthorityLinkChunkKafkaHandler kafkaHandler;
 
-  @Value("${srs.kafka.AuthorityLinkChunkConsumer.loadLimit:2}")
-  private int loadLimit;
+  private static final int AUTHORITY_LINK_CHUNK_CONSUMER_LOAD_LIMIT = 1;
```
Aliaksandr-Fedasiuk, line 21: I don't agree with this solution as I don't see the point in it. We always use only 1 worker to process events from Kafka, and parallel processing of records is performed by multiple instances of the module on prod envs.
`new DeploymentOptions().setWorker(true).setInstances(instancesNumber.orElse(1))`

Aliaksandr-Fedasiuk (changes requested):

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityLinkChunkKafkaHandler.java
```diff
@@ -87,13 +87,14 @@ public AuthorityLinkChunkKafkaHandler(RecordService recordService, KafkaConfig k
 
   @Override
   public Future<String> handle(KafkaConsumerRecord<String, String> consumerRecord) {
-    LOGGER.trace("handle:: Handling kafka record: {}", consumerRecord);
```
Aliaksandr-Fedasiuk, line 90: Please return this log message.

Aliaksandr-Fedasiuk (changes requested):

mod-source-record-storage-server/src/main/java/org/folio/verticle/consumers/AuthorityLinkChunkConsumersVerticle.java
mukhiddin-yusuf: @Aliaksandr-Fedasiuk , The PR is in Draft state now and not ready for review yet, I have already changed the implementation and about to update the PR and make it ready for review soon.

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityLinkChunkKafkaHandler.java
mukhiddin-yusuf: This change was debugging purpose and it would be reverted at the end.

PR #655, opened 2024-11-21
MODSOURCE-824: Endpoint /batch/parsed-records/fetch does not return deleted records

VRohach:
## Purpose
The main goal is to extend _POST /batch/parsed-records/fetch_ endpoint to be able to retrieve DELETED records.
There was added a brand new property- **"includeDeleted"** to the body request to include DELETED records if this set to **"true"**.
If  **"includeDeleted": "false" or  "includeDeleted"  not exists - it will return JUST ACTUAL record.**

## Approach
Additional condition added to DAO based on  "includeDeleted"  property.
Test added.
NEWS updated.

## Learning
JIRA: https://folio-org.atlassian.net/browse/MODSOURCE-824


VRohach:
This PR should be merged only after merging: https://github.com/folio-org/data-import-raml-storage/pull/315

dmytrokrutii (approved):

JavokhirAbdullayev (approved):

RuslanLavrov (approved):

PR #658, opened 2024-12-16
MODSOURCE-803: DuplicateException is thrown when saving batch of Bib records for linked Authority

Aliaksandr-Fedasiuk:
## Purpose
DuplicateException sometimes occurs when updating mark-authority entries linked to marc-bib.

## Approach
Each change goes by a separate query with a unique snapshotId. The query used in this method controlled the date of the operation relative to the snapshotId, and, accordingly, the generation of the record in the records_lb table. Current changes will allow to avoid DuplicateException, but in the future, it is necessary to change the logic so that all incoming requests related to one operation will go with one (equals) snapshotId.

#### TODOS and Open Questions
- [ ] Check logging.
## Learning
[MODSOURCE-803](https://folio-org.atlassian.net/browse/MODSOURCE-803)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
@@ -11,7 +11,6 @@ import static org.folio.dao.util.RecordDaoUtil.ensureRecordHasId;
 import static org.folio.dao.util.RecordDaoUtil.ensureRecordHasSuppressDiscovery;
 import static org.folio.dao.util.RecordDaoUtil.filterRecordByExternalHridValuesWithQualifier;
 import static org.folio.dao.util.RecordDaoUtil.filterRecordByState;
-import static org.folio.dao.util.RecordDaoUtil.getExternalIdType;
 import static org.folio.dao.util.RecordDaoUtil.getExternalIdsConditionWithQualifier;
 import static org.folio.dao.util.SnapshotDaoUtil.SNAPSHOT_NOT_FOUND_TEMPLATE;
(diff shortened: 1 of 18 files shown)


KaterynaSenchenko (approved):

JavokhirAbdullayev (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -173,21 +175,43 @@ public Future<Record> saveRecord(Record record, Map<String, String> okapiHeaders
...
+          ? recordDao.saveRecords(recordCollection, okapiHeaders)
+          : Future.failedFuture(ar.cause()))
+        .recover(err -> retrySave(err, recordCollection, retryCount, maxRetryCount, setMatchedIdsFutures, okapiHeaders));
+    } else {
+      return Future.failedFuture(error);
+    }
+  }
+
   @Override
   public Future<RecordsBatchResponse> saveRecordsByExternalIds(List<String> externalIds,
```
mukhiddin-yusuf, line 216: updating the marc-bibs for the received linked authority event is performed by this ```saveRecordsByExternalIds``` method now. Is there also a need to introduce retry mechanism for ```DuplicateEventException``` here?

Aliaksandr-Fedasiuk: Yes. This case handles the case of changing references via marc-authority DI with an update profile

mukhiddin-yusuf (approved):

RuslanLavrov (approved):

PR #660, opened 2025-01-06
MODDATAIMP-1133 Second update of the same MARC authority / MARC holdings record completes with errors

JavokhirAbdullayev:
## Purpose
[MODDATAIMP-1133](https://folio-org.atlassian.net/browse/MODDATAIMP-1133) Second update of the same MARC authority / MARC holdings record completes with errors

## Approach
fix bug in code


Aliaksandr-Fedasiuk (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/match/AbstractMarcMatchEventHandler.java
```diff
@@ -179,7 +179,7 @@ private Filter.ComparisonPartType prepareComparisonPartType(MatchDetail matchDet
   private Future<RecordCollection> processDefaultMatchField(MatchField matchField, List<String> matchedRecordIds, String tenantId) {
     Condition condition = filterRecordByMultipleIds(matchedRecordIds);
```
RomanChernetskyi, line 180: could we create condition in a single line, like: 
filterRecordByMultipleIds(matchedRecordIds).and(filterRecordByState(Record.State.ACTUAL.value()))

RomanChernetskyi (approved):

PR #662, opened 2025-01-16
fix(marc-fields-order): Fix bib fields order on link update event

viacheslavkol:
## Purpose
Fix bib fields order on link update event

## Approach
- Handle only authority natural id change
- Put uncontroller alphabetic fields before controlled number fields

#### TODOS and Open Questions
- [ ] Check logging.

## Learning
[MODSOURCE-842](https://folio-org.atlassian.net/browse/MODSOURCE-842)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/links/UpdateLinkProcessor.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/links/UpdateLinkProcessor.java
@@ -21,9 +21,14 @@ public class UpdateLinkProcessor implements LinkProcessor {
   @Override
   public Collection<Subfield> process(String fieldCode, List<org.folio.rest.jaxrs.model.Subfield> subfieldsChanges,
                                       List<Subfield> oldSubfields) {
+    if (isOnlyNaturalIdChanged(subfieldsChanges)) {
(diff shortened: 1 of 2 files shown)


mukhiddin-yusuf (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/handlers/links/UpdateLinkProcessor.java
```diff
@@ -21,9 +21,14 @@ public class UpdateLinkProcessor implements LinkProcessor {
...
                                       List<Subfield> oldSubfields) {
+    if (isOnlyNaturalIdChanged(subfieldsChanges)) {
+      return processOnlyNaturalIdChange(oldSubfields, subfieldsChanges.get(0).getValue());
+    }
+
     var authorityIdSubfield = getAuthorityIdSubfield(oldSubfields);
     var authorityNaturalIdSubfield = getAuthorityNaturalIdSubfield(oldSubfields);
-    var controllableSubfields = new ArrayList<Subfield>();
+    var controllableAlphaSubfields = new ArrayList<Subfield>();
+    var controllableDigitSubfields = new ArrayList<Subfield>();
```
psmagin, line 31: I thought it was agreed to have this separation only for uncontrollable sufields. But controllable subfields should have same order as authority record.

psmagin (approved):

PR #664, opened 2025-01-21
[MODSOURCE-840] Extend domain event with source MARC

RomanChernetskyi:
## Purpose
Create/Update/Delete operations in the module should generate related events with the new/old body for the entity (instances/authorities) 
Acceptance Criteria:
- The domain event has new/old fields
- The domain has a unique identifier for the message and a timestamp of the event

## Approach
Send domain events with old and new record

#### Jira
https://folio-org.atlassian.net/browse/MODSOURCE-840

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
@@ -14,7 +14,6 @@ import org.apache.logging.log4j.Logger;
 import org.folio.dao.util.IdType;
 import org.folio.dao.util.RecordType;
 import org.folio.kafka.AsyncRecordHandler;
-import org.folio.rest.jooq.enums.RecordState;
 import org.folio.services.RecordService;
 import org.folio.services.util.KafkaUtil;
 import org.springframework.stereotype.Component;
@@ -75,7 +74,7 @@ public class AuthorityDomainKafkaHandler implements AsyncRecordHandler<String, S
   }
 
   private Future<String> performHardDelete(String authorityId, Map<String, String> okapiHeaders) {
-    return recordService.deleteRecordsByExternalId(authorityId, IdType.RECORD, okapiHeaders).map(authorityId);
+    return recordService.deleteRecordsByExternalId(authorityId, okapiHeaders).map(authorityId);
   }
 
   private void logError(String authorityId, EventSubType subType, String tenantId) {
--- a/mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
@@ -1,13 +1,17 @@
 package org.folio.consumers;
 
 import static org.folio.dao.util.RecordDaoUtil.filterRecordByExternalId;
+import static org.folio.services.util.EventHandlingUtil.toOkapiHeaders;
 
 import io.vertx.core.Future;
 import io.vertx.core.json.JsonObject;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
 import java.util.Collections;
+import java.util.Map;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.folio.dao.util.IdType;
 import org.folio.dao.util.RecordType;
 import org.folio.kafka.AsyncRecordHandler;
 import org.folio.rest.jooq.enums.RecordState;
@@ -34,6 +38,10 @@ public class AuthorityDomainKafkaHandler implements AsyncRecordHandler<String, S
   @Override
   public Future<String> handle(KafkaConsumerRecord<String, String> consumerRecord) {
     log.trace("handle:: Handling kafka record: '{}'", consumerRecord);
+
+    var kafkaHeaders = consumerRecord.headers();
+    var okapiHeaders = toOkapiHeaders(kafkaHeaders);
+
     String authorityId = consumerRecord.key();
     if (isUnexpectedDomainEvent(consumerRecord)) {
       log.trace("handle:: Expected only {} domain type. Skipping authority domain kafka record [ID: '{}']",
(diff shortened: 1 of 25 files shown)


mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -1282,26 +1281,54 @@ public Future<ParsedRecord> updateParsedRecord(Record record, Map<String, String
...
-      )).onSuccess(updated -> recordDomainEventPublisher.publishRecordUpdated(record, okapiHeaders))
-      .map(res -> record.getParsedRecord()));
+    return getQueryExecutor(tenantId).transaction(txQE -> getRecordById(txQE, record.getId())
+      .compose(optionalRecord -> optionalRecord.map(oldRecord ->
+          updateExternalIdsForRecord(txQE, record).compose(updatedRecord ->
+            ParsedRecordDaoUtil.update(txQE, record.getParsedRecord(), ParsedRecordDaoUtil.toRecordType(record))
+              .onSuccess(v -> recordDomainEventPublisher.publishRecordUpdated(oldRecord,
+                updatedRecord
+                  .withErrorRecord(oldRecord.getErrorRecord())
+                  .withRawRecord(oldRecord.getRawRecord())
```
pkjacob, line 1291: Is this correct? Here we assign `oldRecord.getRawRecord()` to `updatedRecord`. Same comment on previous line `.withErrorRecord(oldRecord.getErrorRecord())`

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -1510,8 +1531,11 @@ private MarcBibCollection toMarcBibCollection(QueryResult result) {
   @Override
   public Future<Record> saveUpdatedRecord(ReactiveClassicGenericQueryExecutor txQE, Record newRecord, Record oldRecord, Map<String, String> okapiHeaders) {
     LOG.trace("saveUpdatedRecord:: Saving updated record {}", newRecord.getId());
-    return insertOrUpdateRecord(txQE, oldRecord).compose(r -> insertOrUpdateRecord(txQE, newRecord))
-      .onSuccess(r -> recordDomainEventPublisher.publishRecordUpdated(r, okapiHeaders));
+    return getRecordById(txQE, oldRecord.getId())
+      .compose(optionalRecord -> optionalRecord.map(exisitngRecord ->
+          insertOrUpdateRecord(txQE, oldRecord).compose(r -> insertOrUpdateRecord(txQE, newRecord))
+            .onSuccess(updatedRecord -> recordDomainEventPublisher.publishRecordUpdated(exisitngRecord, updatedRecord, okapiHeaders)))
```
pkjacob, line 1537: typo in `exisitngRecord`. Should be `existngRecord`

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -1510,8 +1531,11 @@ private MarcBibCollection toMarcBibCollection(QueryResult result) {
...
+        .orElse(Future.failedFuture(new NotFoundException(format(RECORD_NOT_FOUND_TEMPLATE, oldRecord.getId())))));
```
pkjacob, line 1538: Per the existing logic, we insert new record to DB if the record do not already exist. Per the new logic, new throw NotFoundException when record do not exist in DB. Is it ok? (I don't know as I am not familiar with this module)

pkjacob:
Hi @RomanChernetskyi, Changes looks good to me except few minor comments. This change is going to break linked-data application. Is it possible to wait few days for merging this PR? We will create a ticket & make changes in linked data per the new event payload.

Here is my understanding about the payload change:

Before:
--
`eventPayload` property of the event contains the impacted `Record` object

After this change:
--
`eventPayload` property will contain a `old` and `new` property. Each of these properties will contain a `Record` object. For SOURCE_RECORD_CREATED event, `old` will contain null

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
RomanChernetskyi: Yes, it is correct, but I already removed this piece of code because we decided not to send raw records and error records to make the domain event less overwhelmed with unneeded data

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
RomanChernetskyi: Fixed, thanks

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
RomanChernetskyi: It's ok because this method updates the existing `oldRecord` and creates `newRecord`, so we expect to have `oldRecord` at DB before executing insert operation

mod-source-record-storage-server/src/main/java/org/folio/consumers/AuthorityDomainKafkaHandler.java
```diff
@@ -48,7 +55,7 @@ public Future<String> handle(KafkaConsumerRecord<String, String> consumerRecord)
     logInput(authorityId, eventSubType, tenantId);
     return (switch (eventSubType) {
       case SOFT_DELETE -> performSoftDelete(authorityId, tenantId);
```
psmagin, line 57: Shouldn't it be a delete event on soft delete?

pkjacob:
@RomanChernetskyi ,

Here is the PR to do the corresponding change in mod-linked-data: https://github.com/folio-org/mod-linked-data/pull/111

pkjacob:
@RomanChernetskyi  Here is the corresponding change in mod-linked-data https://github.com/folio-org/mod-linked-data/pull/111

psmagin (approved):

mod-source-record-storage-server/src/test/java/org/folio/services/RecordServiceTest.java
```diff
@@ -2399,4 +2564,13 @@ private void compareAdditionalInfo(TestContext context, AdditionalInfo expected,
   private void compareExternalIdsHolder(TestContext context, ExternalIdsHolder expected, ExternalIdsHolder actual) {
     context.assertEquals(expected.getInstanceId(), actual.getInstanceId());
   }
+
+  private static <T> T clone(T obj, Class<T> type) {
```
RuslanLavrov, line 2568: it seems we could use the TestUtil.clone method here that we already use in this class or MarcUtil.clone method, because they are identical

JavokhirAbdullayev (approved):

PR #668, opened 2025-01-29
[MODSOURCE-854] Use deleteRecordBy id for soft delete of authority

RomanChernetskyi:
## Purpose
Update soft delete logic for Marc Authority

## Approach
Use deleteRecordBy id for soft delete of authority

#### Jira
https://folio-org.atlassian.net/browse/MODSOURCE-854

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
@@ -1,27 +1,29 @@
 package org.folio.services.handlers.actions;
 
+import io.vertx.core.Future;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.ActionProfile;
 import org.folio.DataImportEventPayload;
-import org.folio.dao.util.RecordType;
+import org.folio.dao.util.IdType;
 import org.folio.processing.events.services.handler.EventHandler;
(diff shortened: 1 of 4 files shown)


mod-source-record-storage-server/src/main/java/org/folio/services/handlers/actions/AbstractDeleteEventHandler.java
```diff
@@ -66,8 +68,10 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload p
   /* Handles DELETE action  */
   private void handlePayload(DataImportEventPayload payload, CompletableFuture<DataImportEventPayload> future) {
     var payloadRecord = Json.decodeValue(payload.getContext().get(getRecordKey()), Record.class);
+    var okapiHeaders = toOkapiHeaders(payload);
     LOG.info("handlePayload:: Handling 'delete' event for the record id = {}", payloadRecord.getId());
-    recordService.updateRecordsState(payloadRecord.getMatchedId(), RecordState.DELETED, RecordType.MARC_AUTHORITY, payload.getTenant())
+    recordService.deleteRecordById(payloadRecord.getMatchedId(), IdType.RECORD, okapiHeaders)
+      .recover(throwable -> throwable instanceof NotFoundException ? Future.succeededFuture() : Future.failedFuture(throwable))
```
VRohach, line 74: @RomanChernetskyi  Let's log here smth. At the debug level - that record was not found (with specific Id). And warn level - with an error message and cause.

Aliaksandr-Fedasiuk (approved):

JavokhirAbdullayev (approved):

dmytrokrutii (approved):

psmagin (approved):

PR #669, opened 2025-01-31
MODSOURCE-855: Error (No space left on device) during CALL delete_old_marc_indexers_versions

Aliaksandr-Fedasiuk:
## Purpose
The first run of the process takes a very long time, and frequent startup puts a heavy load on the processor.

## Approach
- The scheduled start parameter has been added: srs.marcIndexers.delete.plannedTime
- Periodic start parameter: srs.marcIndexers.delete.interval.seconds = 1800
- Added parameter to limit the number of records deleted at a time: srs.marcIndexers.delete.dirtyBatchSize = 100`000
To start a scheduled execution, specify plannedTime like 00:00 or a comma-separated list of values is supported: 01:00,02:30

## Learning
[MODSOURCE-855](https://folio-org.atlassian.net/browse/MODSOURCE-855)

### DIFF
--- a/mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
@@ -45,7 +45,7 @@ public class MarcIndexersVersionDeletionVerticle extends AbstractVerticle {
   @Value("${srs.marcIndexers.delete.interval.seconds:1800}")
   private int interval;
 
-  @Value("${srs.marcIndexers.delete.plannedTime}")
+  @Value("${srs.marcIndexers.delete.plannedTime:}")
   private String plannedTime;
 
   @Value("${srs.marcIndexers.delete.dirtyBatchSize:100000}")
--- a/mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
@@ -60,6 +60,7 @@ public class MarcIndexersVersionDeletionVerticle extends AbstractVerticle {
 
   @Override
   public void start(Promise<Void> startFuture) {
+    LOGGER.info("Specified values: planned time: {}, interval: {}", plannedTime, interval);
     long intervalMillis = interval * 1000L;
 
     if (!"00:00".equals(plannedTime) && interval == 1800) {
--- a/mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
+++ b/mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
@@ -16,6 +16,7 @@ import org.springframework.stereotype.Component;
 import java.time.LocalTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.List;
@@ -56,10 +57,25 @@ public class MarcIndexersVersionDeletionVerticle extends AbstractVerticle {
 
   @Override
   public void start(Promise<Void> startFuture) {
-    this.scheduleTimes = Arrays.stream(deleteTimes.split(","))
-      .map(LocalTime::parse)
-      .sorted()
-      .collect(Collectors.toList());
+    LOGGER.info("deleteOldMarcIndexerVersions:: schedule time: {}", deleteTimes);
+    try {
+      scheduleTimes = Arrays.stream(deleteTimes.split(","))
+        .map(String::trim)
+        .map(time -> {
+          try {
+            return LocalTime.parse(time);
+          } catch (DateTimeParseException e) {
+            LOGGER.error("Error parsing time: '{}'. Defaulting to 01:00", time, e);
+            return LocalTime.of(1, 0);
+          }
+        })
+        .sorted()
+        .collect(Collectors.toList());
+      LOGGER.info("Scheduled times: {}", scheduleTimes);
+    } catch (Exception e) {
+      LOGGER.error("Unexpected error occurred while setting up scheduled times, defaulting to 01:00", e);
+      scheduleTimes = Arrays.asList(LocalTime.of(1, 0));
+    }
     scheduleNextTask(vertx, this::deleteTask);
     startFuture.complete();
   }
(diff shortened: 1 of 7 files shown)


mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
```diff
@@ -25,36 +34,72 @@ public class MarcIndexersVersionDeletionVerticle extends AbstractVerticle {
...
-    });
+    this.scheduleTimes = Arrays.stream(deleteTimes.split(","))
+      .map(LocalTime::parse)
+      .sorted()
+      .collect(Collectors.toList());
+    scheduleNextTask(vertx, this::deleteTask);
     startFuture.complete();
   }
 
+  private void scheduleNextTask(Vertx vertx, Runnable task) {
```
okolawole-ebsco, line 67: I am thinking a bunch of this code could be replaced by doing Vertx.periodicTimer at the start time. Something to consider.

Aliaksandr-Fedasiuk: I think it would be better to move this processing to night time and we can set the time using commas like: srs.marcIndexers.delete.time: 01:00,02:00

mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-5.10.0/2024-12-10--20-00-create-delete-old-marc-indexers-versions.xml
```diff
@@ -31,45 +31,56 @@
...
-        SELECT mrt.marc_id, mrt.version
-        FROM marc_records_tracking mrt
-        WHERE mrt.is_dirty = true
-      ),
-      deleted_dirty AS (
-      DELETE FROM marc_indexers mi
-        USING dirty_records dr
-      WHERE mi.marc_id = dr.marc_id AND dr.version > mi.version
-        RETURNING mi.marc_id
+        WITH dirty_records AS (
```
okolawole-ebsco, line 41: please add `WITH dirty_records AS materialized`. This will allow postgres to compute the CTE before using it elsewhere. It will let the postgres optimizer know if the number of rows in the CTE. When the CTE has 0 rows, a more appropriate query plan will be used with the marc_indexers table.

mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
okolawole-ebsco: I am not disputing moving to specific time. I am saying after starting a specific time, using Vertx.periodic with 1 day duration would be less code. Looking at the code, I can't tell if it works since a lot of the api that is not commonly used by me. This is not a mandatory change, just a suggestion.

mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
```diff
@@ -25,36 +34,72 @@ public class MarcIndexersVersionDeletionVerticle extends AbstractVerticle {
...
+
+  private Future<Boolean> currentDeletion = Future.succeededFuture();
 
   @Autowired
   public MarcIndexersVersionDeletionVerticle(RecordDao recordDao, TenantDataProvider tenantDataProvider) {
     this.recordDao = recordDao;
     this.tenantDataProvider = tenantDataProvider;
   }
 
-  @Value("${srs.marcIndexers.delete.interval.seconds:1800}")
```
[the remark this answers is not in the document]
Aliaksandr-Fedasiuk: hmm... for this I should use two different methods: setPeriodic or setTimer

mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-5.10.0/2024-12-10--20-00-create-delete-old-marc-indexers-versions.xml
Aliaksandr-Fedasiuk: completed, thank you

mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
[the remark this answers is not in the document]
okolawole-ebsco: I agree.

mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
```diff
@@ -25,36 +35,87 @@ public class MarcIndexersVersionDeletionVerticle extends AbstractVerticle {
...
+
+  private final RecordDao recordDao;
+  private final TenantDataProvider tenantDataProvider;
+  private List<LocalTime> scheduleTimes;
+
+  @Value("${srs.marcIndexers.delete.time:01:00}")
+  private String deleteTimes;
+
+  @Value("${srs.marcIndexers.delete.oneTimeLimit:100000}")
+  private Integer oneTimeLimit;
```
okolawole-ebsco, line 48: I think we should rename `oneTimeLimit` to `dirtyBatchSize`.

Aliaksandr-Fedasiuk: Completed. Thank you

mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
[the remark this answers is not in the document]
Aliaksandr-Fedasiuk: Completed, thank you

mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java
```diff
@@ -25,55 +35,122 @@ public class MarcIndexersVersionDeletionVerticle extends AbstractVerticle {
...
-      if (currentDeletion.isComplete()) {
-        currentDeletion = deleteOldMarcIndexerVersions();
-      } else {
-        LOGGER.info("Previous marc_indexers old version deletion still ongoing");
-      }
-    });
+    LOGGER.info("Specified values: planned time: {}, interval: {}", plannedTime, interval);
+    long intervalMillis = interval * 1000L;
+
+    if (!"00:00".equals(plannedTime) && interval == 1800) {
```
okolawole-ebsco, line 66: I think you should remove the default value from the planned time. If planned time is not set use the interval. The problem I see is that if the interval default is changed above, a developer might not know to come change `1800`.

Aliaksandr-Fedasiuk: Fixed. Thank you

okolawole-ebsco (approved):
Looks good!

KaterynaSenchenko (approved):

JavokhirAbdullayev (approved):

VRohach (approved):

RuslanLavrov (approved):

PR #673, opened 2025-02-18
MODSOURCE-859: Create new API POST /source-storage/records/{id}/un-delete

AndreiBordak:
## Purpose
Undelete feature for records

## Approach
API

#### TODOS and Open Questions
- [x] Check logging.
## Learning

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,11 @@
 * [MODSOURCE-860](https://folio-org.atlassian.net/browse/MODSOURCE-860) "Numerics only" option of existing record section does not work during MARC-BIB to MARC-BIB matching 
 * [MODINV-1114](https://folio-org.atlassian.net/browse/MODINV-1114) Extend matching records endpoint to support multiple marc-bib match results processing
 * [MODSOURCE-863](https://folio-org.atlassian.net/browse/MODSOURCE-863) Add index to speed up the querying of a composite record
+* [MODSOURCE-859](https://folio-org.atlassian.net/browse/MODSOURCE-859) Added record undelete endpoint
+
+| METHOD | URL                                      | DESCRIPTION                                         |
+|--------|------------------------------------------|-----------------------------------------------------|
+| POST   | /source-storage/records/{id}/un-delete   | Undelete the record by setting the state to ACTUAL. |
 
 ## 2024-10-28 5.9.0
 * [MODSOURCE-767](https://folio-org.atlassian.net/browse/MODSOURCE-767) Single record overlay creates duplicate OCLC#/035
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,9 +12,9 @@
 * [MODSOURCE-863](https://folio-org.atlassian.net/browse/MODSOURCE-863) Add index to speed up the querying of a composite record
 * [MODSOURCE-859](https://folio-org.atlassian.net/browse/MODSOURCE-859) Added record undelete endpoint
 
-| METHOD | URL                                      | DESCRIPTION                                         |
-|--------|------------------------------------------|-----------------------------------------------------|
-| POST   | /source-storage/records/{id}/un-delete   | Undelete the record by setting the state to ACTUAL. |
+| METHOD | URL                                      | DESCRIPTION         |
+|--------|------------------------------------------|---------------------|
+| POST   | /source-storage/records/{id}/un-delete   | Undelete a record.  |
 
 ## 2024-10-28 5.9.0
 * [MODSOURCE-767](https://folio-org.atlassian.net/browse/MODSOURCE-767) Single record overlay creates duplicate OCLC#/035
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -103,6 +103,15 @@
         },
         {
           "methods": [
+            "POST"
+          ],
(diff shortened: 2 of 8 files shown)


NEWS.md
```diff
@@ -10,6 +10,11 @@
 * [MODSOURCE-860](https://folio-org.atlassian.net/browse/MODSOURCE-860) "Numerics only" option of existing record section does not work during MARC-BIB to MARC-BIB matching 
 * [MODINV-1114](https://folio-org.atlassian.net/browse/MODINV-1114) Extend matching records endpoint to support multiple marc-bib match results processing
 * [MODSOURCE-863](https://folio-org.atlassian.net/browse/MODSOURCE-863) Add index to speed up the querying of a composite record
+* [MODSOURCE-859](https://folio-org.atlassian.net/browse/MODSOURCE-859) Added record undelete endpoint
+
+| METHOD | URL                                      | DESCRIPTION                                         |
+|--------|------------------------------------------|-----------------------------------------------------|
+| POST   | /source-storage/records/{id}/un-delete   | Undelete the record by setting the state to ACTUAL. |
```
pkjacob, line 17: Just `Undelete a record` is enough I think

descriptors/ModuleDescriptor-template.json
```diff
@@ -101,6 +101,15 @@
             "source-storage.records.generation.item.put"
           ]
         },
+        {
```
pkjacob, line 104: I think we have to increment the minor version in line 57 as we are adding a new API

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -421,6 +422,21 @@ public Future<Void> deleteRecordById(String id, IdType idType, Map<String, Strin
...
 
+  @Override
+  public Future<Void> unDeleteRecordById(String id, IdType idType, Map<String, String> okapiHeaders) {
+    var tenantId = okapiHeaders.get(OKAPI_TENANT_HEADER);
+    return recordDao.getRecordByExternalId(id, idType, tenantId)
+      .map(recordOptional -> recordOptional.orElseThrow(() -> new NotFoundException(format(NOT_FOUND_MESSAGE, Record.class.getSimpleName(), id))))
+      .map(record -> {
+        update005field(record);
+        record.withState(Record.State.ACTUAL);
+        record.setAdditionalInfo(record.getAdditionalInfo().withDeleted(false));
```
pkjacob, line 433: See my comment in https://github.com/folio-org/data-import-raml-storage/pull/317.
Do we really need the "deleted" flag in additionalInfo as it is a duplicate information (leader already has deleted flag)

If we decide to add the "deleted" flag in additionalInfo, then we have to set its value as true in the deleteRecordById method.  (line 417)

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -421,6 +422,21 @@ public Future<Void> deleteRecordById(String id, IdType idType, Map<String, Strin
       .compose(record -> updateRecord(record, okapiHeaders)).map(r -> null);
   }
 
+  @Override
+  public Future<Void> unDeleteRecordById(String id, IdType idType, Map<String, String> okapiHeaders) {
```
psmagin, line 426: Does this method send domain event?

pkjacob:
@AndreiBordak,

More code coverage is needed. Changes in RecordServiceImpl is not tested.

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
AndreiBordak: removed redundant initialization

descriptors/ModuleDescriptor-template.json
AndreiBordak: done.

NEWS.md
AndreiBordak: done.

askhat-abishev:
@AndreiBordak there are 3 New isuues

pkjacob (approved):

KaterynaSenchenko (approved):

mod-source-record-storage-server/src/main/java/org/folio/services/RecordServiceImpl.java
```diff
@@ -97,6 +97,7 @@ public class RecordServiceImpl implements RecordService {
   private static final String RECORD_WITH_GIVEN_MATCHED_ID_NOT_FOUND = "Record with given matched id (%s) not found";
   private static final String NOT_FOUND_MESSAGE = "%s with id '%s' was not found";
   private static final Character DELETED_LEADER_RECORD_STATUS = 'd';
+  private static final Character CORRECTED_LEADER_RECORD_STATUS = 'c';
```
RomanChernetskyi, line 100: Could we name it something like ACTUAL_LEADER_RECORD_STATUS

pkjacob: Hi @RomanChernetskyi
Per https://www.loc.gov/marc/bibliographic/bdleader.html, 'c' means 'Corrected or revised'. So, `CORRECTED_LEADER_RECORD_STATUS` is better in my opinion.

PR #675, opened 2025-02-21
MODSOURMAN-1278: Data import job stuck in progress with 99-100%

Aliaksandr-Fedasiuk:
## Purpose
 Bulk edit job was not completed due to Data import job stuck in progress with 100%

## Approach
In the code we added a check for an exception occurring during record processing.

## Learning
[MODSOURMAN-1278](https://folio-org.atlassian.net/browse/MODSOURMAN-1278)


dmytrokrutii (approved):

JavokhirAbdullayev (approved):

KaterynaSenchenko (approved):

RuslanLavrov (approved):

mod-source-record-storage-server/src/main/java/org/folio/consumers/DataImportKafkaHandler.java
```diff
@@ -76,34 +105,63 @@ public Future<String> handle(KafkaConsumerRecord<String, byte[]> targetRecord) {
...
   }
+
+  private static void printLogInfo(Level level, String message, String... args) {
+    printLogInfo(level, message, null, args);
+  }
+
+  private static void printLogInfo(Level level, String message, Throwable tr, String... args) {
+    String formattedMessage;
+    if (tr != null) {
+      formattedMessage = format(ERROR_LOG_MESSAGE_TEMPLATE, message, args[0], args[1], args[2], args[3], tr.getMessage());
```
RuslanLavrov, line 161: propose to pass the whole exception to log instead of the message so that we could also have a stack trace in case an investigation is needed

Aliaksandr-Fedasiuk: Completed. Thanks!

PR #682, opened 2025-03-13
Eureka ECS bugfest - 504 Gateway Time-out Error is returned for MARC Search request against Central tenant

VRohach:
## Purpose
The main purpose - increase performance for the /source-storage/stream/marc-record-identifiers POST endpoint.

## Approach
Function immutable_to_date(text) updated, and now using in the search SQL request, new index was added.

## Learning
JIRA: https://folio-org.atlassian.net/browse/MODSOURCE-871

### DIFF
--- a/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-5.10.0/2025-12-03--15-00-create-composite-index-and-update-immutable-to-date-function.xml
+++ b/mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-5.10.0/2025-12-03--15-00-create-composite-index-and-update-immutable-to-date-function.xml
@@ -7,18 +7,20 @@
 
   <changeSet id="2025-12-03--15-00--update-immutable-to-date-function" author="Volodymyr_Rohach">
     <sql splitStatements="false">
-      CREATE OR REPLACE FUNCTION ${database.defaultSchemaName}.immutable_to_date(input_text TEXT) RETURNS date AS $$
+      CREATE OR REPLACE FUNCTION ${database.defaultSchemaName}.immutable_to_date(input_text TEXT)
+      RETURNS date AS $$
       DECLARE
       output_date DATE;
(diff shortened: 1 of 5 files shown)


mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-5.10.0/2025-12-03--15-00-create-composite-index-and-update-immutable-to-date-function.xml
```diff
@@ -0,0 +1,32 @@
...
+      END;
+      RETURN output_date;
+      END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+    </sql>
+  </changeSet>
+
+  <changeSet id="2025-12-03--15-00-composite-index-field-no-and-immutable-function" author="Volodymyr_Rohach">
+    <sql>
+      CREATE INDEX IF NOT EXISTS idx_marc_indexers_005_field_no_and_immutable_to_date ON ${database.defaultSchemaName}.marc_indexers_005 (field_no, ${database.defaultSchemaName}.immutable_to_date(value));
```
Aliaksandr-Fedasiuk, line 29: I think we should remove the unused idx_marc_indexers_005_value_date index.

Aliaksandr-Fedasiuk (changes requested):

mod-source-record-storage-server/src/main/resources/liquibase/tenant/scripts/v-5.10.0/2025-12-03--15-00-create-composite-index-and-update-immutable-to-date-function.xml
VRohach: Good point!
Done. Review it please

dmytrokrutii (approved):

RuslanLavrov (approved):

PR #688, opened 2025-04-30
Prepare v5.10.5 development iteration

SvitlanaKovalova1:
## Purpose
Release v5.10.5, prepare v5.10.6 development iteration

## Approach
[MODSOURCE-903](https://folio-org.atlassian.net/browse/MODSOURCE-903)


mukhiddin-yusuf (approved):

natali-zaitseva (approved):

RuslanLavrov (approved):

RuslanLavrov:
it looks like that the line separator has been changed from LF to CRLF in the pom.xml files, and that's why the difference is shown for all lines in the pom.xml files

PR #725, opened 2026-02-19
MODSOURCE-936 - Upgrade module to Vert.x 5.0

RuslanLavrov:
## Purpose
upgrade the module to Vertx 5

## Approach
* The `vertx-jooq-classic-reactive` has been removed due to it does not support Vert.x 5, and the ReactiveClassicGenericQueryExecutor has been replaced with PgPoolQueryExecutor
* The retry logic for transaction operations has been moved to the PgPoolQueryExecutor
* The Pump class was deprecated and removed in Vert.x 5 so ReadStream.pipe() has been used instead in the SourceStorageStreamImpl


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

mod-source-record-storage-server/src/main/java/org/folio/dao/RecordDaoImpl.java
```diff
@@ -679,85 +648,75 @@ private RecordsIdentifiersCollection toRecordsIdentifiersCollection(QueryResult
 
   @Override
   public Future<Optional<Record>> getRecordById(String id, String tenantId) {
-    return getQueryExecutor(tenantId).transaction(txQE -> getRecordById(txQE, id));
+    return getQueryExecutor(tenantId).transaction(queryExecutor -> getRecordById(getQueryExecutor(tenantId), id));
```
okolawole-ebsco, line 651: This is running in the transaction
`queryExecutor -> getRecordById(getQueryExecutor(tenantId), id)`
It should be:
`queryExecutor -> getRecordById(queryExecutor, id)`

You can see from other changes in this file that the `queryExecutor` should be passed instead of creating a new one.

This is causing multiple connections for requests to get a record. The first query executor retrieves a connection and starts a transaction. But then, because we acquire a second query executor, all of the database actions happen on the second connection. The default connection pool size is 15. So if there were 8 concurrent calls to get a record, the 8th call will not be able to get 2 connections and hence a deadlock.

When a deadlock happens, the 8th concurrent call times out when getting a connection. Vert.x 5 does not show any stacktrace so we are not seeing any logs. https://github.com/eclipse-vertx/vertx-sql-client/blob/3fe39b4ebb3238ee4f4a16147d604fe0f4f09f62/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/pool/SqlConnectionPool.java#L208 
`VertxException("Timeout waiting for connection", noStackTrace=true)`

RuslanLavrov: thank you, I've found this place as well, but was trying to reproduce proof that it causes transactions to hang. I didn't guess that it can cause transactions deadlock. I think deadlock occurs when 15 calls to this method managed to get connection to open transaction, but get stuck waiting for connection to execute query on non-transactional executor because no free connections are left.

RuslanLavrov: Olamide, I've found that, fortunately, in fact, the code that you provided by the link fails handler and eventually, Pool returns failed future in case of timeout. This is available since Vertx v5.0.6. Currently, SRS uses Vertx v5.0.5, which does not return failed future in case of timeout. That's why the transactions get stuck forever, instead of failing due to timeout, and there are no error logs about timeout while executing query on non-transactional executor. It looks like Vertx v5.0.5 still can return NoStackTraceThrowable("Timeout") while obtaining connection using Pool.transaction() method, but does not during execution query on the Pool itself (e.g. using Pool.preparedQuery()).

